PR #15, opened 2017-04-06
utils.go generation from vendor.yml

mcarmonaa:


mcarmonaa:
TestIsVendor has been changed because it didn't test vendored files and directories.

internal/code-generator/assets/utils.go.tmpl
```diff
@@ -0,0 +1,84 @@
+package slinguist
```
mcuadros: instead of having all the code in the template, we can have only the regexp and the code in a real go file

PR #22, opened 2017-04-19
Added language detection by shebang functionality

mcarmonaa:

PR #23, opened 2017-04-19
Added language detection by filename strategy

mcarmonaa:
This PR must be merged after https://github.com/src-d/simple-linguist/pull/22


filename.go
```diff
@@ -0,0 +1,10 @@
+package slinguist
+
+func GetLanguageByFilename(filename string) (lang string, safe bool) {
```
alcortesm: Add a comment here.

alcortesm: Explain why safe here means a different thing than in the Shebang part.

filename.go
```diff
@@ -0,0 +1,10 @@
...
+	lang, safe = languagesByFilename[filename]
+	if lang == "" {
+		lang = OtherLanguage
```
alcortesm: The role of OtherLanguage is pretty important, maybe we should explain that clearly somewhere.

alcortesm: Can we make OtherLanguage the zero value of a lang type to get rid of the extra code to handle OtherLanguage?

filename.go
mcarmonaa: safe is used as a part of the process to fall through the different strategies to detect the language used in a source file. Maybe safe would be described in other place of the code (GetLanguage function, README?) and in other PR.

filename.go
mcarmonaa: Maybe this can be performed in another PR.

filename.go
alcortesm: :+1:

filename.go
alcortesm: :+1:

common.go
```diff
@@ -36,6 +36,14 @@ func GetLanguageExtensions(language string) []string {
 
 // GetLanguage return the Language for a given filename and file content.
 func GetLanguage(filename string, content []byte) string {
+	if lang, safe := GetLanguageByFilename(filename); safe {
+		return lang
+	}
+
+	if lang, safe := GetLanguageByShebang(content); safe {
```
bzz: Isn't this and the rest of [shebang strategy](https://github.com/src-d/simple-linguist/pull/23/files#diff-a9c3f74e11f5aefc6762b6c5f6e78401R19) supposed to be part of #22 ?

mcarmonaa: Yes it is, this it the reason because https://github.com/src-d/simple-linguist/pull/22 must be merged before this PR. This PR is branched from the other PR, from commit f45efec to back the commits are the same that in https://github.com/src-d/simple-linguist/pull/22

PR #28, opened 2017-05-25
Added Classifier as the last step in detection strategies

mcarmonaa:
Classifier function is not taking in a list of candidates in the strategies execution yet.


classifier.go
```diff
@@ -0,0 +1,97 @@
...
+	scores := classifier.Classify(content, languages)
+	language := getLangugeHigherScore(scores)
+	return language
+}
+
+func getLangugeHigherScore(scores map[string]float64) string {
+	var language string
+	higher := -math.MaxFloat64
+	for lang, score := range scores {
+		higher = math.Max(higher, score)
```
ajnavarro: **Comment:** Why not just do:
```go
if higher < score {
    language = lang
    higher = score
}
```

classifier.go
```diff
@@ -0,0 +1,97 @@
...
+	languages := make([]string, 0, len(candidates))
+	for _, candidate := range candidates {
+		lang := GetLanguageByAlias(candidate)
+		if lang != OtherLanguage {
+			languages = append(languages, lang)
+		}
+	}
+
+	scores := classifier.Classify(content, languages)
+	language := getLangugeHigherScore(scores)
```
ajnavarro: **Comment:** You can return directly the function result:
```go
return getLangugeHigherScore(scores)
```

mcarmonaa: Ok!

classifier.go
mcarmonaa: I thought to accuraly compare float64 it was better option use this function. Should I change it anyway?

mcarmonaa:
@ajnavarro I performed the changes about clone linguist for test, could you take a look?

classifier.go
ajnavarro: Yep, is more acurately, but I don't know if here we need this amount of accuracy. Here a test: https://play.golang.org/p/58w93fQ_cU

classifier.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"math"
+
+	"gopkg.in/src-d/simple-linguist.v1/internal/tokenizer"
+)
+
+// GetLanguageByClassifier takes in a content and a list of candidates, and apply the classifier's Classify method to
+// get the most probably language. If classifier is null then DefaultClassfier will be used.
+func GetLanguageByClassifier(content []byte, candidates []string, classifier Classifier) string {
+	if len(content) == 0 {
```
ajnavarro: I think this logic should be part of the classifier.

classifier.go
```diff
@@ -0,0 +1,96 @@
...
+		return OtherLanguage
+	}
+
+	if classifier == nil {
+		classifier = DefaultClassifier
+	}
+
+	languages := make([]string, 0, len(candidates))
+	for _, candidate := range candidates {
+		lang := GetLanguageByAlias(candidate)
```
ajnavarro: Maybe a better interface for `string GetLanguageByAlias(string)` could be `string,bool GetLanguageByAlias(string)`. Doing this, you don't need to return and check by "OtherLanguage".

internal/code-generator/generator/generator.go
```diff
@@ -21,6 +21,14 @@ func FromFile(fileToParse, outPath, tmplPath, tmplName, commit string, generate
 		return err
 	}
 
+	if err := formatedWrite(outPath, source); err != nil {
```
ajnavarro: just:
```go
return  formatedWrite(outPath, source)
```

internal/code-generator/generator/samplesfreq.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return err
+	}
+
+	buf := &bytes.Buffer{}
+	if err := executeFrequenciesTemplate(buf, freqs, frequenciesTmplPath, frequenciesTmplName, commit); err != nil {
+		return err
+	}
+
+	if err := formatedWrite(outPath, buf.Bytes()); err != nil {
```
ajnavarro: just return:
```go
return formatedWrite(outPath, buf.Bytes())
```

internal/code-generator/generator/samplesfreq.go
```diff
@@ -0,0 +1,202 @@
...
+		TokensTotal:    tokensTotal,
+		LanguageTotal:  languageTotal,
+		Tokens:         tokens,
+		LanguageTokens: languageTokens,
+		Languages:      languages,
+	}, nil
+}
+
+func getSamples(samplesDir string, langDir os.FileInfo) ([]string, error) {
+	const subDir = "filenames"
```
ajnavarro: this constant should be outside this function?

PR #30, opened 2017-06-02
Improvement rearranged code

mcarmonaa:
The following improvements has been addressed:
* rearranged code and removed files
* commented public API
* fixed some tests in internal/code-generator/generator
* changed test to use github.com/stretchr/testify
* tokenizer reads up no more than bytesLimit bytes (100000)
* strategies are now public functions which returns []string of languages
* the "old public strategies" are now exposed functions that use internally the strategies.
* Classifier interface's method Classify now takes in candidates as a map[string]float64
* now the classifier is taking in a map of candidates in the sequence strategies' execution, and is only executed if the map of candidates is not empty.
* simple-linguist cli now generates the same output as linguist --json

Ready for version 2?


common.go
```diff
@@ -1,71 +1,171 @@
...
+func GetLanguage(filename string, content []byte) string {
+	candidates := make([]string, 0, 5)
+	for _, strategy := range strategies {
+		languages := strategy(filename, content)
+		if len(languages) == 1 {
+			return languages[0]
+		}
+
+		if len(languages) > 0 {
+			candidates = append(candidates, languages...)
```
abeaumont: there may be duplicated languages in candidates. Is that ok? shouldn't a set be better here?

common.go
```diff
@@ -1,71 +1,171 @@
...
+func GetLanguageByShebang(content []byte) (lang string, safe bool) {
+	languages := getLanguagesByShebang("", content)
+	if len(languages) == 0 {
+		lang = OtherLanguage
+		return
 	}
 
-	if lang, safe := GetLanguageByFilename(filename); safe {
-		return lang
+	lang = languages[0]
```
abeaumont: This code is equivalent to the code in `GetLanguageByFilename`, right? If so, be consistent with the style.
It may be even better to extract the pattern of shared code to a private function, to avoid the unnecessary duplication.

common.go
```diff
@@ -1,71 +1,171 @@
...
+// OtherLanguage is used as a zero value when a function can not return a specific language.
 const OtherLanguage = "Other"
 
-var (
-	ExtensionsByLanguage map[string][]string
-	ignoredExtensions    = map[string]bool{
-		".asc": true, ".cgi": true, ".fcgi": true, ".gml": true, ".fx": true,
-		".vhost": true,
+var strategies = []func(filename string, content []byte) (languages []string){
+	getLanguagesByModeline,
```
mcuadros: why we made the strategies private? I prefer having `strategies` public and also the functions allowing to the user choose his own strategies if they want.

common.go
mcarmonaa: You are right, it doesn't affect to overcome but process twice the same candidate. Maybe the Classify interface must takes in a map[language]weight, with weight as an int that stores the number of times the candidate has been seen. The weight could be used in an alternative classifier that wants consider a pre-conditions dynamically (like the number of times the candidates appeared in the chain of strategies). What do you think @smola ?

PR #32, opened 2017-06-08
Subpackage for generated files

dpaz:
frequencies.go and type.go are still being generated in the slinguist package due to a cyclic dependency problem.


mcarmonaa:
classifier type contains  `languagesLogProbabilities` `tokensLogProbabilities` `tokensTotal`, so you must generate this fields in frequencies.go and then assign them to the DefaultClassfier declaring it in classifier.go.

As for type.go, maybe generating `languagesType` as a map[string]int and then casting in a variable declared in the top level... Anyway it seems that add unnecessary complexity.

Is it necessary put generated files in a subpackage that doesn't export functionality only gather maps? Maybe to see at a glance what are the generated files is enough renaming those files to generated_whatever.go or something like that.

smola:
@dpaz This needs rebase.

internal/code-generator/generator/generator_test.go
```diff
@@ -212,7 +212,7 @@ func (g *GeneratorTestSuite) TestGenerationFiles() {
 
 		outPath, err := ioutil.TempFile("/tmp", "generator-test-")
 		assert.NoError(g.T(), err)
-		defer os.Remove(outPath.Name())
+		//defer os.Remove(outPath.Name())
```
mcarmonaa: Uncomment this

common.go
```diff
@@ -397,28 +405,28 @@ func GetLanguagesBySpecificClassifier(content []byte, candidates []string, class
...
 type Type int
 
 // Type's values.
 const (
-	Unknown Type = iota
-	Data
-	Programming
-	Markup
-	Prose
+	Unknown     = 0
```
smola: These should be `Type = iota` as before.

common.go
```diff
@@ -397,15 +405,15 @@ func GetLanguagesBySpecificClassifier(content []byte, candidates []string, class
...
+	return data.ExtensionsByLanguage[language]
 }
 
 // Type represent language's type. Either data, programming, markup, prose, or unknown.
 type Type int
 
 // Type's values.
 const (
-	Unknown Type = iota
+	Unknown = iota
```
smola: What about `Type`? Why is it not typed?

smola:
Please, rebase.

PR #34, opened 2017-06-12
Improved detection and changed strategies' signature

mcarmonaa:
The following changes has been addressed:
* added linguist/samples/filenames to map
* improved extension detection
* changed regexp for empty files in content.go generation
* fixed bug in classifier
* changed strategies' signature
* Classify interface now returns []string
* all strategies are in common.go (removed shebang.go and modeline.go)
* changed TestLinguistCorpus to use GetLanguage and print corner cases. Now the test fails if a language is not detected properly.
* internal/code-generator now use generator.File as a function type for all source files generated
* test modifications due to the performed changes (internal/code-generator/generator test now uses a fixed linguist commit to get the inputs for the tests)
* GetLanguage is now GetLanguage(...) (language, safe) and GetLanguages(...) (languages []string)

I've kept functions that returns one language and safe because they could be usefull for API clients (for examples we use GetLanguageByExtension(*) (language, safe) in utils.go)

We differ from linguist: [**DETECTION ISSUES!**](https://github.com/src-d/enry/issues/35)


classifier.go
```diff
@@ -10,7 +10,7 @@ import (
 // The scores must order the candidates so as the highest score be the most probably language of the content. The candidates is
```
abeaumont: s/The candidates/Candidates/

classifier.go
```diff
@@ -10,7 +10,7 @@ import (
...
 // a map which can be used to assign weights to languages dynamically.
 type Classifier interface {
-	Classify(content []byte, candidates map[string]float64) map[string]float64
+	Classify(content []byte, candidates map[string]float64) (scores map[float64]string)
```
abeaumont: A map with a float64 key looks really awkward. Is really a map from float to strings what you want? Or just an slice with language and weight pairs or similar?

mcarmonaa: Yes it's awkward, I built the map[float64]string to can sort the list of languages by their scores without additionally data structures. I moved the sort logic to the classifier and now the Classifier interface returns a []string of languages.

abeaumont: I don't think it's much harder to do it properly (even if you have to write a comparator function), and it should certainly be more efficient. OTOH, using a map will break as soon as you have 2 languages with the same weight.

ajnavarro:
@mcarmonaa review this amount of files is difficult, could you split this PR, at least, in two? per example, you can do a PR changing the package name to `enry`, this will reduce the amount of file changes into this PR.

mcarmonaa:
@ajnavarro  i squashed commits after rename to enry to a previous commit, so there are changes mixed :S, my fault. From now on I'll be careful to do smaller PR and better split.

Anyway some files don't need review:
* alias.go, content.go, documentation.go, extension.go, filename.go, frequencies.go, interpreter.go, type.go, vendor.go are generated files. Changes in this files come from changes in linguist's files we parse.
* under internal/code-generator/assets are modified templates to renamed to enry
* inside internal/code-generator/generator/tests_files are files used as a reference for test, are generated files from a specific commit from linguist
* shebang.go and modeline.go has been removed and their content it's located now in common.go

ajnavarro:
@mcarmonaa Anyways, you can do a new PR changing the package name from scratch, and then, when it will be merged, rebase it from here.

mcarmonaa:
This PR must be merged after https://github.com/src-d/enry/pull/36

mcuadros:
@mcarmonaa rebase

mcarmonaa:
@smola split function GetLanguage is done

smola:
@mcarmonaa Do not add more things to this PR please.

common.go
```diff
@@ -25,39 +25,20 @@ var DefaultStrategies = []Strategy{
...
-
-		if len(languages) > 0 {
-			candidates = append(candidates, languages...)
-		}
-	}
-
-	return firstLanguage(languages)
+// to find out the most probably language to return. If there are more than one possibles languages
+// it returns safe to false.
+func GetLanguage(filename string, content []byte) (language string, safe bool) {
```
smola: Sorry, I think I didn't express it properly: maintain previous signature, we don't want that `safe` boolean, what we want is adding `GetLanguages` with the language list.

PR #36, opened 2017-06-13
renamed package and cli to enry

mcarmonaa:
* package statements and cli now use the new name "enry"

### DIFF
--- a/cli/enry/main.go
+++ b/cli/enry/main.go
@@ -44,8 +44,8 @@ func main() {
 			relativePath = relativePath + "/"
 		}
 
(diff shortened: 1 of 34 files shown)


cli/enry/main.go
```diff
@@ -9,7 +9,7 @@ import (
 	"os"
 	"path/filepath"
 
-	"gopkg.in/src-d/enry.v1"
+	enry "gopkg.in/src-d/enry.v1"
```
smola: you should be able to remove the alias here.

mcarmonaa: done

PR #42, opened 2017-06-16
avoid return code is non-zero in case of warnings

dpaz:


cli/enry/main.go
```diff
@@ -80,9 +80,6 @@ func main() {
 	data, _ := json.MarshalIndent(out, "", "  ")
 	fmt.Printf("%s\n", data)
```
mcarmonaa: Remove this empty line

Issue #47, opened 2017-06-21
executed test should match commit used to generate the data


mcuadros:
The CI is failing due to using the master branch of `linguist` with a code generated with another commit of `linguist`

mcarmonaa:
@mcuadros Should we run tests against the same linguist's commit our files were generated or let it as a way to know linguist has been updated and we must generate files again?

mcuadros:
Yep, we should run the test with the same version

mcarmonaa:
Tests are failing because the linguist version in local differs from the linguist's commit our files were generated. TestLinguistCorpus() walk throw the .linguist/samples and takes the expected language from the directory name that contains all files related to a language.

Because linguist name for Perl 6 parsed from .linguist/lib/linguist/languages.yml has changed from "Perl6" to "Perl 6", the expected language in the test is "Perl 6", but with the files generated from a previous linguist's commit, Perl 6 in our map of languages is still parsed as "Perl6"

To run against the same commit our files were generated, we can generate another file that contains only a private variable with the linguist's commit all files have been generated:
`var linguistCommit= "b6460f8ed6b249281ada099ca28bd8f1230b8892"`

So we can have documented at a glance from what commit files were generated (although all generated files have a comment with this commit), and we can use this variable to download linguist to a temp directory, do a checkout to this commit and run tests like in internal/code-generator/generator's tests is done.

PR #49, opened 2017-06-21
fixed linguist's commit which is used by tests

mcarmonaa:
fixed issue https://github.com/src-d/enry/issues/47

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
@@ -11,15 +12,46 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
-type SimpleLinguistTestSuite struct {
+const linguistURL = "https://github.com/github/linguist.git"
+
+type EnryTestSuite struct {
 	suite.Suite
+	repoLinguist string
 }
 
-func TestSimpleLinguistTestSuite(t *testing.T) {
-	suite.Run(t, new(SimpleLinguistTestSuite))
+func TestEnryTestSuite(t *testing.T) {
+	suite.Run(t, new(EnryTestSuite))
 }
 
-func (s *SimpleLinguistTestSuite) TestGetLanguage() {
+func (s *EnryTestSuite) SetupSuite() {
+	var err error
+	s.repoLinguist, err = ioutil.TempDir("", "linguist-")
+	assert.NoError(s.T(), err)
+
+	cmd := exec.Command("git", "clone", linguistURL, s.repoLinguist)
+	err = cmd.Run()
+	assert.NoError(s.T(), err)
+
+	cwd, err := os.Getwd()
+	assert.NoError(s.T(), err)
+
+	err = os.Chdir(s.repoLinguist)
+	assert.NoError(s.T(), err)
+
+	cmd = exec.Command("git", "checkout", linguistCommit)
+	err = cmd.Run()
+	assert.NoError(s.T(), err)
+
+	err = os.Chdir(cwd)
+	assert.NoError(s.T(), err)
+}
+
+func (s *EnryTestSuite) TearDownSuite() {
+	err := os.RemoveAll(s.repoLinguist)
+	assert.NoError(s.T(), err)
+}
+
+func (s *EnryTestSuite) TestGetLanguage() {
 	tests := []struct {
 		name     string
 		filename string
@@ -38,7 +70,7 @@ func (s *SimpleLinguistTestSuite) TestGetLanguage() {
 	}
 }
 
-func (s *SimpleLinguistTestSuite) TestGetLanguagesByModelineLinguist() {
+func (s *EnryTestSuite) TestGetLanguagesByModelineLinguist() {
 	const (
 		modelinesDir = ".linguist/test/fixtures/Data/Modelines"
 		samplesDir   = ".linguist/samples"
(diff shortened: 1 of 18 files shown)


common_test.go
```diff
@@ -4,22 +4,54 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 )
 
-type SimpleLinguistTestSuite struct {
+const lingustURL = "https://github.com/github/linguist.git"
+
+type EnryTestSuite struct {
 	suite.Suite
+	tmpLinguist string
```
alcortesm: According to the use we are giving to this variable, I would call it localLinguistRepo or something similar.

common_test.go
```diff
@@ -4,22 +4,54 @@ import (
...
+	"os/exec"
 	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/suite"
 )
 
-type SimpleLinguistTestSuite struct {
+const lingustURL = "https://github.com/github/linguist.git"
```
alcortesm: s/lingustURL/linguistURL/

common_test.go
```diff
@@ -4,22 +4,54 @@ import (
...
 	suite.Suite
+	tmpLinguist string
+}
+
+func TestEnryTestSuite(t *testing.T) {
+	suite.Run(t, new(EnryTestSuite))
+}
+
+func (e *EnryTestSuite) SetupSuite() {
+	tmpLinguist, err := ioutil.TempDir("", "linguist-")
```
alcortesm: assign to `e.tmpLinguist` here directly and use it on subsequent calls instead of creating a new variable `tmpLinguist`.

common_test.go
```diff
@@ -4,22 +4,54 @@ import (
...
+	assert.NoError(e.T(), err)
+	e.tmpLinguist = tmpLinguist
+
+	cmd := exec.Command("git", "clone", lingustURL, tmpLinguist)
```
alcortesm: This is weird, looks like an integration test to me.

We shouldn't be downloading anything for the tests, instead, we should include a fixture in this repository that we can use to test our behaviour, independently of what's written in a certain commit of the official linguist repository.

But then generated code is complicated and maybe it is too difficult. What's the actual situation here?

alcortesm: I mean, can we just use a very small and representative corpus as a fixture, generate whatever you want to generate from it and test it, instead of downloading a whole repo to a local dir and then checking out a specific version of that repo?

mcarmonaa: Actually it is like an integration test.

To generate the files, we have already cloned linguist in the project's directory. Maybe it's better perform a checkout over this repo, do the test and then come back to the master of this repo.

Should I do something like this?

alcortesm: > Actually it is like an integration test.

That was unexpected, I would expect that to be done from the outside, maybe through a test script from the makefile or something. But if you are OK with it, i'm OK with it :).

> To generate the files, we have already cloned linguist in the project's directory. Maybe it's better perform a checkout over this repo, do the test and then come back to the master of this repo.

Instead of relying on it being there, first check if the information is already there, if it is, do the required checkout, if not clone and checkout.

I am not sure about the part of checking out master again after the tests, sorry, I'm not very familiar with the internals of enry.  I will approve the PR now and rely on the reviews of wiser reviewers :).

bzz:
Looks good. It may be a bit easier to review if a generated code would in a separate commit from the code, written for for any other purpose.

mcarmonaa:
@bzz all the code in this PR is related to fix  https://github.com/src-d/enry/issues/47, generated code in this PR is a new generated file that it wasn't before in the project.

abeaumont:
@mcarmonaa Even if it's newly generated code it can easily be separated in a commit and helps the reviewer know the difference in advance :)

utils_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func (s *SimpleLinguistTestSuite) TestIsVendor() {
+func (e *EnryTestSuite) TestIsVendor() {
```
abeaumont: same thing about `ts` name instead of `e` as above

internal/code-generator/generator/linguist-commit.go
```diff
@@ -0,0 +1,17 @@
+package generator
+
+import (
+	"bytes"
+	"text/template"
+)
+
+// Commit takes in commit and builds source file from tmplPath. It's comply with type File signature.
```
abeaumont: * s/`Commit takes in commit and builds source file`/`Commit takes a commit and builds the source file`/?
* `It's comply` sounds incorrect. Maybe you meant 'It complies'?

common_test.go
```diff
@@ -4,22 +4,54 @@ import (
...
+type EnryTestSuite struct {
 	suite.Suite
+	repoLinguist string
+}
+
+func TestEnryTestSuite(t *testing.T) {
+	suite.Run(t, new(EnryTestSuite))
+}
+
+func (e *EnryTestSuite) SetupSuite() {
```
abeaumont: `ts` is probably a better name for a testsuite, and wouldn't have required this name change because project name's changed

utils_test.go
smola: we usually use `s` for every `*Suite`

internal/code-generator/generator/linguist-commit.go
```diff
@@ -0,0 +1,17 @@
+package generator
+
+import (
+	"bytes"
+	"text/template"
+)
+
+// Commit takes in commit and builds source file from tmplPath. It complies with type File signature.
```
abeaumont: This still sounds a bit weird to me _Commit takes **a** commit and builds **the** source file from tmplPath_ sounds better.

mcarmonaa: fixed

PR #50, opened 2017-06-23
Gitatributes

dpaz:
Added support for .gitattributes which works similar to [github/linguist](https://github.com/github/linguist)


README.md
```diff
@@ -44,6 +45,36 @@ To run the tests
     make test
 
 
+.gitAttributes
```
smola: It's `.gitattributes`, all lower case.

utils.go
```diff
@@ -66,3 +81,55 @@ func IsBinary(data []byte) bool {
...
 	return true
 }
+
+func loadGitattributes() (map[string]string, error) {
+	gitAttributes := map[string]string{}
+	data, err := ioutil.ReadFile(".gitattributes")
+	if err != nil {
+		if err != os.ErrNotExist {
+			log.Println(".gitattributes: " + err.Error())
+		}
```
mcarmonaa: Add an empty line after

utils.go
```diff
@@ -66,3 +81,55 @@ func IsBinary(data []byte) bool {
...
+		return nil, err
+	}
+
+	if data != nil {
```
mcarmonaa: If you actually want to avoid the code inside the if statement, you should check for `len(data) > 0`, if ioutil.ReadFile doesn't fail it could return an empty slice and an empty slice in not nil.

utils.go
```diff
@@ -66,3 +81,55 @@ func IsBinary(data []byte) bool {
...
+		tokens := strings.Fields(string(data))
+		for i := 0; i < len(tokens); i = i + 2 {
```
mcarmonaa: Maybe you should use a scanner or split before by lines, it's not so clear what are you doing here at a glance

utils.go
```diff
@@ -66,3 +81,55 @@ func IsBinary(data []byte) bool {
...
+			gitAttributes[tokens[i]] = tokens[i+1]
+		}
+	}
+
+	return gitAttributes, nil
+}
+
+func parseAttributes(attributes map[string]string) {
+	for key, val := range attributes {
+		switch {
```
mcarmonaa: what happens if the same path has several attributes?

utils.go
```diff
@@ -66,3 +81,55 @@ func IsBinary(data []byte) bool {
...
+	tokens := strings.Split(attribute, "=")
+	if len(tokens) == 2 {
+		regExp, err := regexp.Compile(regExpString)
+		if err == nil {
+			languageGitAttributes[regExp] = tokens[1]
+		}
+	}
+}
+
+func LoadGitAttributes() {
```
mcarmonaa: put this function before loadGitattributes() and add a comment if it is a public function

common.go
```diff
@@ -87,6 +88,12 @@ func GetLanguageByClassifier(content []byte, candidates []string) (language stri
 	return getLanguageByStrategy(GetLanguagesByClassifier, "", content, candidates)
 }
 
+// GetLanguageByGitAttributes returns the language assigned for a given regular expresion in .gitattributes.
+// This strategy needs to be initialized callin LoadGitAttributes
```
mcarmonaa: s/callin/calling/

README.md
```diff
@@ -33,6 +33,7 @@ fmt.Println(lang)
 lang := enry.GetLanguage("foo.cpp", "<cpp-code>")
 ```
 
+
 Developmemt
 -----------
 *enry* re-uses parts of original [linguist](https://github.com/github/linguist) especially data in `languages.yml` to generate internal data structures. In oreder to update to latest upstream run
```
mcarmonaa: s/oreder/order/

mcarmonaa:
@dpaz  You should add tests for new functionality

utils.go
```diff
@@ -66,3 +81,55 @@ func IsBinary(data []byte) bool {
...
+		}
+	}
+}
+
+func processLanguageAttr(regExpString string, attribute string) {
+	tokens := strings.Split(attribute, "=")
+	if len(tokens) == 2 {
+		regExp, err := regexp.Compile(regExpString)
+		if err == nil {
+			languageGitAttributes[regExp] = tokens[1]
```
mcarmonaa: you should pass tokens[1] to GetLanguageByAlias() before store it in the map, this ensures the language will be stored like a "standard name" if it is a known language for enry

common.go
alcortesm: Can we do this on demand instead of having to call `LoadGitAttributes`?

alcortesm: What is the argument to this function? a filename or a regular expression (function signature says the former, comment explains the later)

common.go
```diff
@@ -434,3 +441,20 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
 
 	return
 }
+
+// GetLanguagesByGitAttributes returns the language assigned in .gitattributes if the regular expresion
+// matchs with the filename
```
alcortesm: Misleading comment: It returns a slice of languages, not a language.

The difference between this method and `GetLanguageByGitAttributes` should be made much clear and probably their names should be much different also.

alcortesm: Why is this function ignoring most of its arguments?

If it is to implemement and interface, it would be nice to know, and why are they being ignored in this particular implementation. Also give ignored arguments a good name, like `_`?

common.go
```diff
@@ -434,3 +441,20 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
...
+func GetLanguagesByGitAttributes(filename string, content []byte, candidates []string) []string {
+	return languageByGitAttribute(filename)
+}
+
+func languageByGitAttribute(filename string) []string {
```
alcortesm: Why does this function exists? can its code be part of its caller body?
Shouldn't it be called `languages...` instead of `language...`?

common.go
dpaz: I added some info to say that it is implementing an interface, the rest of the things that you are telling me are like this to be coherent with the other strategies that are written similarly.

README.md
```diff
@@ -44,6 +45,36 @@ To run the tests
     make test
 
 
+.gitAttributes
+--------------
+
+Like in linguist you can override the strategies via `.gitattributes` file.
+Add a `.gitattributes` file to the directory and use the enry matchers `enry-documentation`,`enry-language` or `enry-vendored` to do the override.
```
abeaumont: I don't know how `linguist` works, but why are we using specific enry matchers instead of reusing the linguist ones? Or are we adding these enry matchers _on top of_ the linguist ones?

utils.go
```diff
@@ -66,3 +83,99 @@ func IsBinary(data []byte) bool {
 
 	return true
 }
+
+// LoadGitattributes reads and parse the file .gitattributes wich overrides the standards strategies
```
abeaumont: s/parse/parses/
s/wich/which/
s/standards/standard/

utils.go
```diff
@@ -66,3 +83,99 @@ func IsBinary(data []byte) bool {
...
+func LoadGitattributes() {
+	rawAttributes, err := loadGitattributes(".gitattributes")
+	if err == nil && len(rawAttributes) > 0 {
+		parseAttributes(rawAttributes)
+	}
+}
+
+func loadGitattributes(name string) (map[string]string, error) {
```
abeaumont: Ignore me if this is some kind of Go convention, but having 2 functions whose only difference in the name is the case of the first letter seems pretty confunsing to me

utils.go
```diff
@@ -66,3 +83,99 @@ func IsBinary(data []byte) bool {
...
+	gitattributes := map[string]string{}
+	data, err := ioutil.ReadFile(name)
+	if err != nil {
+		if err != os.ErrNotExist {
+			log.Println(".gitattributes: " + err.Error())
```
abeaumont: `log.Println(name + ": " + err.Error())`?

utils_test.go
```diff
@@ -81,6 +82,42 @@ func (s *SimpleLinguistTestSuite) TestIsBinary() {
 	}
 }
 
+func (s *SimpleLinguistTestSuite) TestLoadLine() {
```
abeaumont: I guess this should be `EnryTestSuite` as this point :)

README.md
dpaz: its only a name they called linguist-documentation or linguist-vendored because is part of linguist, I thought that because we are on enry we should call it enry-vendored and enry-documentation.

utils_test.go
dpaz: yes the PR was not rebased till yesterday I will change this

README.md
abeaumont: Yes, but if you name them differently you're forcing the user to have to support both linguist and enry separately. If you just use the same name, enry will already support all those attributes already customized by users for linguist.

common.go
```diff
@@ -434,3 +441,16 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
 
 	return
 }
+
+// GetLanguagesByGitattributes returns a length 1 slice with the language assigned in .gitattributes if the regular expresion
+// matchs with the filename. It is comply with the signature to be a Strategy type.
```
alcortesm: Improve the English and the explanation.

alcortesm: Note that it returns an empty slice if the filename does not match any of the rules in .gitattributes.  Should this return nil instead?

common.go
```diff
@@ -434,3 +441,16 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
...
+func GetLanguagesByGitattributes(filename string, content []byte, candidates []string) []string {
+	languages := []string{}
```
alcortesm: Maybe it is better, to clarify the interface to remove this allocation from here and move it to the return statement, as otherwise we are allocating a slice to return it empty in case the filename does not match any languageGitattributes.

common.go
alcortesm: This should be called `GetLanguagesByGitAttributes` with an uppercase `A`.  The same for all git attributes functions and variables, except for the name of the file (`.gitattributes`) itself, of course.

common.go
```diff
@@ -453,9 +453,11 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
 // GetLanguagesByGitAttributes returns either a string slice with the lenguage if the filename match with a regExp in .gitattributes
```
abeaumont: s/lenguage/language/
s/match/matches/

common.go
```diff
@@ -453,9 +453,11 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
...
 //or return nil in case of none regexp matchs the filename . It complies with the signature to be a Strategy type.
```
abeaumont: * space before _or_
* s/return/returns/
* no space after _filename_
* s/in case of none/in case no/
* s/matchs/matches/

common.go
```diff
@@ -453,9 +453,11 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
...
 func GetLanguagesByGitAttributes(filename string, content []byte, candidates []string) []string {
-	for regExp, language := range languageGitAttributes {
-		if regExp.MatchString(filename) {
-			return []string{language}
+	if loadedGitAttributes != nil {
```
abeaumont: I don't think having a global variable floating around different files is a good idea.

utils.go
```diff
@@ -133,86 +166,97 @@ func loadLine(line string, gitattributes map[string][]string) error {
...
-		err = processVendorAttr(key, attribute)
-	case strings.Contains(attribute, "linguist-documentation"):
-		err = processDocumentationAttr(key, attribute)
-	case strings.Contains(attribute, "linguist-language="):
-		err = processLanguageAttr(key, attribute)
-	default:
-		err = errors.New(fmt.Sprintf("gitattributes: The matcher %s doesn't exists\n", attribute))
-		log.Printf(err.Error())
+func newGitAttributes() *gitAttributes {
+	gA := gitAttributes{
```
abeaumont: `gA` is not too meaningful, and it doesn't make sense to upcase the 'A'.
I'd prefer something like 'attrs' or 'attributes', but since really short names are already used elsewhere, at least use proper casing: `ga`

common.go
```diff
@@ -442,3 +449,17 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
 
 	return
 }
+
+// GetLanguagesByGitAttributes returns either a string slice with the lenguage if the filename match with a regExp in .gitattributes
+//or return nil in case of none regexp matchs the filename . It complies with the signature to be a Strategy type.
+func GetLanguagesByGitAttributes(filename string, content []byte, candidates []string) []string {
+	if loadedGitAttributes != nil {
```
alcortesm: The idiomatic way to do this in Go is to return nil if the loadedGitAttributes is == nil.  That way, the happy path of the function (the for loop) has the smallest level of indentation.  For a better explanation of this concept, see Golang UK Conference 2016 - Mat Ryer - Idiomatic Go Tricks (https://www.youtube.com/watch?v=yeetIgNeIkc).

alcortesm: Sorry, I approved the PR by mistake, let's try again:

The idiomatic way to do this in Go is to return nil if the loadedGitAttributes is == nil. That way, the happy path of the function (the for loop) has the smallest level of indentation. For a better explanation of this concept, see Golang UK Conference 2016 - Mat Ryer - Idiomatic Go Tricks (https://www.youtube.com/watch?v=yeetIgNeIkc).

common.go
```diff
@@ -450,15 +450,20 @@ func GetLanguageByAlias(alias string) (lang string, ok bool) {
...
 
-// GetLanguagesByGitattributes returns a length 1 slice with the language assigned in .gitattributes if the regular expresion
-// matchs with the filename. It is comply with the signature to be a Strategy type.
-func GetLanguagesByGitattributes(filename string, content []byte, candidates []string) []string {
-	languages := []string{}
-	for regExp, language := range languageGitattributes {
-		if regExp.MatchString(filename) {
-			return append(languages, language)
-		}
+// GetLanguagesByGitAttributes returns either a string slice with the language if the filename matches with a regExp in .gitattributes
```
mcarmonaa: put this function with the rest of the strategies

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+type attrType int
+
+const (
+	vendor attrType = iota
+	documentation
+	generated
+	language
+)
+
+const _attrType_name = "vendordocumentationgeneratedlanguage"
```
mcarmonaa: What are indicating the underscores? I though this is not so idiomatic in Go.

On the other hand, why don't you use the previous constants as a string turning attrType a string type? Seems like it's adding unnecessary complexity, with _attr_index and the method for attrType String

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+
+var _attrType_index = [...]uint8{0, 6, 19, 28, 36}
```
mcarmonaa: why the underscores?

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+	boolAttributes   map[attrType]boolAttribute
+	regExpAttributes map[attrType]regExpAttribute
+}
+
+type overrideError struct {
+	attribute attrType
+	path      string
+}
+
+func (e *overrideError) Error() string {
```
mcarmonaa: should we take attributes overriding as an error?

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+	if val, ok := gitAttrs.boolAttributes[vendor].attributes[path]; ok {
+		return val
+	}
+
+	return data.VendorMatchers.Match(path)
+}
+
+// IsVendor returns whether or not path is a documentation path.
+func (gitAttrs *GitAttributes) IsDocumentation(path string) bool {
+	if val, ok := gitAttrs.boolAttributes[documentation].attributes[path]; ok {
```
mcarmonaa: Maybe you can keep only those attributes which are "true". If the attribute is not in the map it's false, so you can check if the attribute is in the map.
`
if _, ok := gitAttrs.boolAttributes[documentation].attributes[path]; ok {
        return ok
}
`

In this way, if we eventually allow attributes overriding, when load an attribute set to false, look for it in the map to delete it

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+		return val
+	}
+
+	return data.DocumentationMatchers.Match(path)
```
mcarmonaa: you should use IsVendor(path)

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+
+// IsVendor returns whether or not path is a generated path.
+func (gitAttrs *GitAttributes) IsGenerated(path string) bool {
+	if val, ok := gitAttrs.boolAttributes[generated].attributes[path]; ok {
+		return val
+	}
+	return false
+}
+
+// GetLanguage get the language of a file matching the langauge attributes given.
```
mcarmonaa: s/langauge/language

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+// Returns either a empty string or the language if the regExp matches
+func (gitAttrs *GitAttributes) GetLanguage(filename string) string {
+	for regExp, language := range gitAttrs.regExpAttributes[language].attributes {
+		if regExp.MatchString(filename) {
+			return language
+		}
+	}
+
+	return ""
```
mcarmonaa: return OtherLanguage

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+// NewGitAttributes initialize a Gitattributes object
+func NewGitAttributes() *GitAttributes {
+	gitAttrs := GitAttributes{
+		boolAttributes: map[attrType]boolAttribute{
+			vendor:        boolAttribute{kind: vendor, matchers: []string{"linguist-vendored", "linguist-vendored=false"}, attributes: map[string]bool{}},
```
mcarmonaa: If attrType were a string, you could remove the "kind" field from boolAttribute and regExpAttribute. In fact, the kind of these structs should be their type. If a type needs to explicitly keep a "kind" maybe it should be several types with perhaps a common parent type

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+func (gitAttrs *GitAttributes) LoadGitAttributes(path string, reader io.Reader) []error {
+	rawAttributes, errArr := loadRawGitAttributes(reader)
+	if len(rawAttributes) == 0 {
+		return []error{}
+	}
+
+	return append(gitAttrs.parseAttributes(path, rawAttributes), errArr...)
+}
+
+func (gitAttrs *GitAttributes) String() string {
```
mcarmonaa: Why this function is needed?

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+	out := ""
+	for key, val := range gitAttrs.boolAttributes {
+		out += fmt.Sprintf("Type: %s	Attributes: %v\n", key, val.attributes)
+	}
+
+	for key, val := range gitAttrs.regExpAttributes {
+		out += fmt.Sprintf("Type: %s	Attributes: %v\n", key, val.attributes)
+	}
```
mcarmonaa: add an empty line after this one

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+		errArr = append(errArr, err)
+		return nil, errArr
+	}
+
+	if len(data) > 0 {
+		lines := strings.Split(string(data), "\n")
+		for _, line := range lines {
+			err := loadLine(line, rawAttributes)
+			if err != nil {
+				errArr = append(errArr, err)
```
mcarmonaa: If we found a malformed .gitattributes file, we can return a "malformed error" and return nil and one error.

I don't know if we should allow to load a malformed .gitattributes file and return a slice of errors. We can be sure that the rest of attributes are ok if there are errors in the file?

gitattributes.go
```diff
@@ -0,0 +1,251 @@
...
+// IsVendor returns whether or not path is a documentation path.
+func (gitAttrs *GitAttributes) IsDocumentation(path string) bool {
+	if val, ok := gitAttrs.boolAttributes[documentation].attributes[path]; ok {
+		return val
+	}
+
+	return data.DocumentationMatchers.Match(path)
+}
+
+// IsVendor returns whether or not path is a generated path.
```
abeaumont: Same here :)

gitattributes.go
dpaz: This is autogenerated code by the command tool stringer its the most simple way to do a toString when you got an enum  [stringer](https://godoc.org/golang.org/x/tools/cmd/stringer)

gitattributes.go
dpaz: Well,  if an error is found nothing happens when the LoadAttributes function ends, it returns a slice of warnings that you can ignore, I'm only using this for traceability and testing purposes

common.go
```diff
@@ -410,6 +410,24 @@ func GetLanguagesBySpecificClassifier(content []byte, candidates []string, class
 	return classifier.Classify(content, mapCandidates)
 }
 
+// GetLanguagesByGitAttributes returns either a string slice with the language if the filename matches with a regExp in .gitattributes
```
abeaumont: These rows are too long

common.go
```diff
@@ -410,6 +410,24 @@ func GetLanguagesBySpecificClassifier(content []byte, candidates []string, class
...
+//or returns a empty slice in case no regexp matches the filename. It complies with the signature to be a Strategy type.
```
abeaumont: space befor 'or'

abeaumont:
@dpaz Apart from the review changes, please rebase.

README.md
```diff
@@ -139,6 +139,36 @@ Using [linguist/samples](https://github.com/github/linguist/tree/master/samples)
 * all files for SQL language fall to the classifier because we don't parse this [disambiguator expresion](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L433) for `*.sql` files right. This expression doesn't comply with the pattern for the rest of [heuristics.rb](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb) file.
 
 
+.gitAttributes
```
abeaumont: all lowercase

README.md
```diff
@@ -139,6 +139,36 @@ Using [linguist/samples](https://github.com/github/linguist/tree/master/samples)
...
+
+Like in linguist you can override the strategies via `.gitattributes` file.
+Add a `.gitattributes` file to the directory and use the same matchers that you would uses in linguist `linguist-documentation`,`linguist-language` or `linguist-vendored` to do the override.
+
+#### Vendored code
+
+Use the `linguist-vendored` attribute to vendor or un-vendor paths.
+
+```
+$cat .gitattributes
```
abeaumont: space before `cat`

README.md
```diff
@@ -139,6 +139,36 @@ Using [linguist/samples](https://github.com/github/linguist/tree/master/samples)
...
+#### Documentation
+
+Documentation works the same way as vendored code but using `linguist-documentation` and `linguist-documentation=false`.
+
+#### Language assignation
+
+If you want some files to be classified according to certain language use `linguist-language=[language]`.
+
+```
+$cat .gitattributes
```
abeaumont: space before `cat`

README.md
```diff
@@ -139,6 +139,36 @@ Using [linguist/samples](https://github.com/github/linguist/tree/master/samples)
...
+.*\.go linguist-language=MyFavouriteLanguage
+```
+Note, that the regular expression that match the file name should be one compatible with go, see: [Golang regexp](https://golang.org/pkg/regexp/).
+>>>>>>> 18e17c9... Added support for language attribute
```
abeaumont: this should not be here

README.md
```diff
@@ -139,6 +139,36 @@ Using [linguist/samples](https://github.com/github/linguist/tree/master/samples)
...
+
+#### Language assignation
+
+If you want some files to be classified according to certain language use `linguist-language=[language]`.
+
+```
+$cat .gitattributes
+.*\.go linguist-language=MyFavouriteLanguage
+```
+Note, that the regular expression that match the file name should be one compatible with go, see: [Golang regexp](https://golang.org/pkg/regexp/).
```
abeaumont: * s/Note, that/Note that/
* s/match/matches/
* s/be one/be/

cli/enry/main.go
```diff
@@ -66,20 +73,18 @@ func main() {
...
-				}
+		content, err := ioutil.ReadFile(path)
+		if err != nil {
+			errors = true
+			log.Println(err)
+			return nil
+		}
+
+		language := gitAttributes.GetLanguage(filepath.Base(path))
+		if len(language) == 0 {
```
abeaumont: `language == enry.OtherLanguage`?

abeaumont:
@mcarmonaa can you validate if everything is ok from your side?

README.md
```diff
@@ -182,6 +185,38 @@ to get time averages for main detection function and strategies for the whole sa
 if you want see measures by sample file
 
 
+
+.gitattributes
+--------------
+
+Like in linguist you can override the strategies via `.gitattributes` file.
+Add a `.gitattributes` file to the directory and use the same matchers that you would uses in linguist `linguist-documentation`,`linguist-language` or `linguist-vendored` to do the override.
```
mcarmonaa: s/uses/use/

README.md
```diff
@@ -182,6 +185,38 @@ to get time averages for main detection function and strategies for the whole sa
...
+
+Documentation works the same way as vendored code but using `linguist-documentation` and `linguist-documentation=false`.
+
+#### Language assignation
+
+If you want some files to be classified according to certain language use `linguist-language=[language]`.
+
+```
+$ cat .gitattributes
+.*\.go linguist-language=MyFavouriteLanguage
```
mcarmonaa: If the regex must be compatible with Golang regexp, does it mean that a .gitattribute file used in linguist has a different syntax? Does it make .gitattributes files for enry incompatibles with linguist ones?

common.go
```diff
@@ -407,6 +414,25 @@ func GetLanguagesBySpecificClassifier(content []byte, candidates []string, class
 	return classifier.Classify(content, mapCandidates)
 }
 
+// GetLanguagesByGitAttributes returns either a string slice with the language
+// if the filename matches with a regExp in .gitattributes or returns a empty slice
```
mcarmonaa: s/a empty/an empty/

mcarmonaa:
As [linguist's README says](https://github.com/github/linguist/blob/master/README.md): `Add a .gitattributes file to your project and use standard git-style path matchers for the files you want to override...`, If we have a different way to specify a path and don't support the same style that *linguist* uses, we won't be compatible.

The same `.gitattributes` file should work with linguist and enry.

dpaz:
Like @mcarmonaa   is saying, I don't really know If go regexps are a superset of git path mathers, If they are this will be fully compatible, if not we need to add some intermediate regexp motor, a converter or something.

abeaumont:
@dpaz how does `linguist` implement those regexps?

vmarkovtsev:
We need a hero to finish this!

hero

PR #51, opened 2017-06-26
remove reflection-based slice sort

erizocosmico:
Remove the reflection-based slice sort and substitute it by a type implementing `sort.Interface`. `SliceStable` uses reflection and as a result is way slower than this. Also, it's only available in Go 1.8 and above (distros like fedora don't have Go 1.8 yet on the official channels).


smola:
Please, rebase.

erizocosmico:
@smola done

PR #67, opened 2017-07-06
check for extension and filename first, without load a file content

mcarmonaa:


cli/enry/main.go
```diff
@@ -61,16 +61,22 @@ func main() {
...
-			log.Println(err)
-			return nil
-		}
-
-		language := enry.GetLanguage(filepath.Base(path), content)
-		if language == enry.OtherLanguage {
-			return nil
+		var language string
+		var ok bool
+		if language, ok = enry.GetLanguageByExtension(path); !ok {
```
abeaumont:
```go
var language string
var ok bool
if language, ok = enry.
```

could just be replaced by
```go
if language, ok := enry.
```

or am I missing something?

mcarmonaa: I need the language variable out of the `if` scope to append it then to the out[] map

abeaumont: Oh, I see. You can still leave the `ok` out then :)

mcarmonaa: Actually I think I can't https://play.golang.org/p/KWI_sbLOaJ

abeaumont: I mean something like

```go
var language string
if language, ok := enry...
````

Comment (mcarmonaa) [inline cli/enry/main.go]: Actually I can't, if I don't define  the `ok` var out the `if` too, the `:=` in the if statement would define two new variables https://play.golang.org/p/KWI_sbLOaJ

Comment (abeaumont) [inline cli/enry/main.go]: Oh, I see :S. Then it makes more sense to just define them outside the if statement:
```go
language, ok := enry..
if !ok {
}
```

Anyway, I'm quite happy with it as it is, just trying to understand go better. Thanks for the explanation!

PR #68, opened 2017-07-06
added cli documentation to README

mcarmonaa:


README.md
```diff
@@ -33,7 +33,57 @@ fmt.Println(lang)
 lang := enry.GetLanguage("foo.cpp", "<cpp-code>")
 ```
 
-Developmemt
+CLI
+-----------------
+You can use enry as a command,
```
abeaumont: minor: a blank between section title and content

README.md
```diff
@@ -33,7 +33,57 @@ fmt.Println(lang)
...
+```bash
+$ enry --help
+enry, A simple (and faster) implementation of github/linguist
+usage: enry <path>
+              enry <path> [--json] [--breakdown]
+              enry [--json] [--breakdown]
+```
+and it will return an output similar to *linguist*'s output,
```
abeaumont: minor: blank lines between code sections and text for better readability

README.md
```diff
@@ -33,7 +33,57 @@ fmt.Println(lang)
...
+
+Go
+parser/main.go
+```
+even the JSON flag,
+```bash
+$ enry --json
+{"Gnuplot":["plot-histogram.gp"],"Go":["parser/main.go"],"Ruby":["linguist-samples.rb","linguist-total.rb"],"Shell":["parse.sh","plot-histogram.sh","run-benchmark.sh","run-slow-benchmark.sh","run.sh"]}
+```
+but the main point here is not the similarity to *linguist* else the difference because **_enry doesn't need a git repository to work!_**
```
abeaumont: This may deserve it's own paragraph, the construction also seems a bit awkward grammatically. Maybe something like:

Note that even if *enry*'s CLI is compatible with *linguist*, its main point is that, contrary to *linguist*, **_enry doesn't need a git repository to work!_**

Issue #72, opened 2017-07-10
Enry used to analize a file


dpaz:
Linguist can analyze a simple file but enry no

PR #73, opened 2017-07-10
Cli to analyze a single file

dpaz:
#72


cli/enry/main.go
```diff
@@ -136,6 +139,50 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
 	}
 }
 
+func printFile(file string, buff *bytes.Buffer) {
```
abeaumont: I think there are some issues with this function as it is (which are not completely independent):
* I don't think `printFile`, as it is, is the proper name for this function since it doesn't print anything.
* Why doesn't it just print the file as promised instead of handling a buffer and doing the printing outside?
* Why is a pointer to a buffer a function parameter instead of just being a return value?

cli/enry/main.go
```diff
@@ -136,6 +139,50 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
...
+	language := enry.GetLanguage(file, content)
+	mimeType := enry.GetMimeType(language)
+
+	buff.WriteString(fmt.Sprintf("%s: %d lines (%d sloc)\n", filepath.Base(file), totalLines, sloc))
+	buff.WriteString(fmt.Sprintf("  type:      %s\n", fileType))
+	buff.WriteString(fmt.Sprintf("  mime type: %s\n", mimeType))
+	buff.WriteString(fmt.Sprintf("  language:  %s\n", language))
+}
+
+func getLines(file string, content []byte) (int, int) {
```
abeaumont: * sloc is a confunsing name, since it's ill-defined (not sure if `sloc` is consistently used in enry to refer to non-blank lines, in that case it'd be ok)
* Messing bytes and strings is confusing, I'd simplify it just using strings.

```go
func getNonBlankLines(file string, content string) (int, int) {
  totalLines := strings.Count(content, "\n")
  blankLines := strings.Count(content, "\n\n")
  return totalLines, totalLines - blankLines
}

totalLines, nonBlackLines := getLines(file, content.String())
```

cli/enry/main.go
```diff
@@ -136,6 +139,50 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
...
+		return "Binary"
+	default:
+		return "Text"
+	}
+}
+
+func isImage(file string) bool {
+	index := strings.LastIndex(file, ".")
+	extension := file[index:]
+	if extension == ".png" || extension == ".jpg" || extension == ".jpeg" || extension == ".gif" {
```
abeaumont: Only this image formats are supported? Why?

cli/enry/main.go
```diff
@@ -136,6 +139,50 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
...
+	case isImage(file):
+		return "Image"
+	case enry.IsBinary(content):
+		return "Binary"
+	default:
+		return "Text"
+	}
+}
+
+func isImage(file string) bool {
```
abeaumont: Wouldn't it make sense to include this function is `utils.go` as the `IsBinary` function?

cli/enry/main.go
dpaz: the isImage func was directly translated from linguist, those are the format that they support

cli/enry/main.go
dpaz: the name and the input parameters are a little weird because I try to keep coherence with the other functions that prints the json etc.

data/mimeType.go
```diff
@@ -0,0 +1,461 @@
+package data
+
+// CODE GENERATED AUTOMATICALLY WITH gopkg.in/src-d/enry.v1/internal/code-generator
+// THIS FILE SHOULD NOT BE EDITED BY HAND
+// Extracted from github/linguist commit: d5c8db3fb91963c4b2762ca2ea2ff7cfac109f68
+
+var LanguagesMime = map[string]string{
+	"1C Enterprise":    "",
```
mcarmonaa: Those languages that don't have a mime type shouldn't appear in the map. Skip this entries when generate the file

data/mimeType.go
```diff
@@ -0,0 +1,461 @@
...
+	"ABAP":             "",
+	"ABNF":             "",
+	"AGS Script":       "text/x-c&#43;&#43;src",
```
mcarmonaa: This string is escaped, you must use text/template instead html/template as import in the file where the function to generate this file is defined

internal/code-generator/generator/mimeType.go
```diff
@@ -0,0 +1,53 @@
+package generator
+
+import (
+	"bytes"
+	"html/template"
```
mcarmonaa: Use text/template instead

utils.go
```diff
@@ -53,6 +53,14 @@ func IsDocumentation(path string) bool {
 	return data.DocumentationMatchers.Match(path)
 }
 
+func GetMimeType(language string) string {
+	if mime, ok := data.LanguagesMime[language]; ok {
```
mcarmonaa: If you keep languages with a string zero value assigned in the map, this code will always return `ok` set to true

cli/enry/main.go
```diff
@@ -139,33 +136,36 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
...
 }
 
-func printFile(file string, buff *bytes.Buffer) {
+func printFileAnalysis(file string) string {
 	content, err := ioutil.ReadFile(file)
 	if err != nil {
 		fmt.Println(err)
 	}
-	totalLines, sloc := getLines(file, content)
+	totalLines, linesOfCode := getLines(file, string(content))
```
abeaumont: Sorry to be picky here, but `linesOfCode` is just the same as `sloc` and equally ill-defined. As in my previous comment, unless a line of code has been unambiguously defined somewhere as non-blank lines for the whole project, it's just an arbitrary decision (the same way that in the text `sloc` should be just called `non-blank`)

utils.go
```diff
@@ -53,6 +53,23 @@ func IsDocumentation(path string) bool {
 	return data.DocumentationMatchers.Match(path)
 }
 
+func IsImage(file string) bool {
+	index := strings.LastIndex(file, ".")
+	extension := file[index:]
+	if extension == ".png" || extension == ".jpg" || extension == ".jpeg" || extension == ".gif" {
+		return true
+	}
+	return false
```
mcarmonaa: add an empty line before this one

cli/enry/main.go
```diff
@@ -136,6 +136,44 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
 	}
 }
 
+func printFileAnalysis(file string) string {
+	content, err := ioutil.ReadFile(file)
+	if err != nil {
+		fmt.Println(err)
+	}
+	totalLines, linesOfCode := getLines(file, string(content))
```
mcarmonaa: add an empty line before this one

cli/enry/main.go
```diff
@@ -136,6 +136,44 @@ func printPercents(out map[string][]string, buff *bytes.Buffer) {
...
+		`%s: %d lines (%d sloc)
+		   type:      %s
+			 mime_type: %s
+			 language:  %s
+		`,
+		filepath.Base(file), totalLines, linesOfCode, fileType, mimeType, language)
+}
+
+func getLines(file string, content string) (int, int) {
+
```
mcarmonaa: remove this empty line

utils.go
```diff
@@ -53,6 +53,23 @@ func IsDocumentation(path string) bool {
 	return data.DocumentationMatchers.Match(path)
 }
 
+func IsImage(file string) bool {
```
mcarmonaa: Maybe it's better call the parameter `path` instead `file` to be consistent with the rest of the functions

utils.go
```diff
@@ -53,6 +53,23 @@ func IsDocumentation(path string) bool {
...
+	index := strings.LastIndex(file, ".")
+	extension := file[index:]
```
mcarmonaa: you can use filepath.Ext(file) to get the extension

utils.go
```diff
@@ -53,20 +53,24 @@ func IsDocumentation(path string) bool {
...
+	extension := filepath.Ext(path)
 	if extension == ".png" || extension == ".jpg" || extension == ".jpeg" || extension == ".gif" {
 		return true
 	}
+
 	return false
 }
 
-func GetMimeType(language string) string {
+func GetMimeType(path string, language string) string {
```
mcarmonaa: You don't need a `path` parameter here

dpaz: Without the path, we can't know if the file is an image

abeaumont:
@dpaz note that tests are failing now

abeaumont:
@mcarmonaa anything pending from your review? lgtm but please confirm.

PR #75, opened 2017-07-11
documented divergences from linguist

mcarmonaa:


README.md
```diff
@@ -126,6 +126,14 @@ To run the tests
     make test
 
 
+Divergences from linguist
+------------
+
+Using [linguist/samples](https://github.com/github/linguist/tree/master/samples) as a set against run tests the following issues were found:
+* with [hello.ms](https://github.com/github/linguist/blob/master/samples/Unix%20Assembly/hello.ms) we can detect the language (Unix Assembly) because we don't have a matcher in contentMatchers (content.go) for Unix Assembly. Linguist use this [regexp](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L300) which we can't port.
```
abeaumont: * s/Linguist use/Linguist uses/
* I don't understand why we can detect correctly a language that `linguist` can't due to us not having their `regexp`... maybe that could be clarified?
* May it be a good idea to put the `regexp` directly here and link to the file instead to a specific line that may be unstable?

README.md
```diff
@@ -126,6 +126,14 @@ To run the tests
...
+
+* all files for SQL language fall to the classifier because we don't parse right this [disambiguator expresion](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb#L433) for "*.sql" files. This expression doesn't comply with the pattern for the rest of [heuristics.rb](https://github.com/github/linguist/blob/master/lib/linguist/heuristics.rb) file.
```
abeaumont: * s/parse right .../parse ... right/
* `*.sql` instead of "*.sql"

README.md
mcarmonaa: We can't detect hello.ms correctly, s/we can detect the/we can't detect/

Issue #74, opened 2017-07-11
release should attach a binary distribution for linux and osx


mcuadros:
In order to be easy to install and use enry, we should provide a precompiled version of it, for linux and osx.

vmarkovtsev:
Any updates on this?

abeaumont:
@vmarkovtsev we lack the correct github token to make this work. CC: @mcuadros.

marnovo:
@mcuadros also need a token. Same as last one.

vmarkovtsev:
Cool! Where do I take them now?

abeaumont:
The fix didn't work, will look at why.

abeaumont:
It seems it was a token permission problem, it works now.

PR #76, opened 2017-07-11
Included src-d/ci/Makefile.main in Makefile

mcarmonaa:
https://github.com/src-d/enry/issues/74

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 .linguist
 benchmarks/output
+.ci
+Makefile.main
--- a/Makefile
+++ b/Makefile
@@ -1,32 +1,36 @@
-COVERAGE_REPORT := coverage.txt
-COVERAGE_PROFILE := profile.out
-COVERAGE_MODE := atomic
+# Package configuration
+PROJECT = enry
+COMMANDS = cli/enry
+
+# Including ci Makefile
+MAKEFILE = Makefile.main
+CI_REPOSITORY = https://github.com/src-d/ci.git
+CI_FOLDER = .ci
+
+# If you need to build more than one dockerfile, you can do so like this:
+# DOCKERFILES = Dockerfile_filename1:repositoryname1 Dockerfile_filename2:repositoryname2 ...
+
+$(MAKEFILE):
+	@git clone --quiet $(CI_REPOSITORY) $(CI_FOLDER); \
+	cp $(CI_FOLDER)/$(MAKEFILE) .;
+
+-include $(MAKEFILE)
 
 LINGUIST_PATH = .linguist
 
 # build CLI
-VERSION := $(shell git describe --tags --abbrev=0)
-COMMIT := $(shell git rev-parse --short HEAD)
-LDFLAGS = -s -X main.Version=$(VERSION) -X main.GitHash=$(COMMIT)
+LOCAL_TAG := $(shell git describe --tags --abbrev=0)
+LOCAL_COMMIT := $(shell git rev-parse --short HEAD)
+LOCAL_BUILD := $(shell date +"%m-%d-%Y_%H_%M_%S")
+LOCAL_LDFLAGS = -s -X main.version=$(LOCAL_TAG) -X main.build=$(LOCAL_BUILD) -X main.commit=$(LOCAL_COMMIT)
 
 $(LINGUIST_PATH):
 	git clone https://github.com/github/linguist.git $@
 
-test: $(LINGUIST_PATH)
-	go test -v ./...
+clean-linguist:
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -1 +1,3 @@
 .linguist
+.ci
+Makefile.main
```
abeaumont: minor: newline at EOF

cli/enry/main.go
```diff
@@ -111,13 +112,13 @@ func main() {
 func usage() {
 	fmt.Fprintf(
 		os.Stderr,
-		`  %[1]s %[2]s commit: %[3]s
+		`  enry %[1]s build: %[2]s commit: %[3]s
   enry, A simple (and faster) implementation of github/linguist
-  usage: %[1]s <path>
-         %[1]s [-json] [-breakdown] <path>
-         %[1]s [-json] [-breakdown]
+  usage: enry <path>
```
abeaumont: what's the reason for hardcoding the binary name now?

Makefile
```diff
@@ -1,39 +1,40 @@
...
 LINGUIST_PATH = .linguist
 
 # build CLI
-VERSION := $(shell git describe --tags --abbrev=0)
-COMMIT := $(shell git rev-parse --short HEAD)
-LDFLAGS = -s -X main.Version=$(VERSION) -X main.GitHash=$(COMMIT)
+LOCAL_TAG := $(shell git describe --tags --abbrev=0)
+LOCAL_COMMIT := $(shell git rev-parse --short HEAD)
+LOCAL_BUILD := $(shell date +"%m-%d-%Y_%H_%M_%S")
+LDFLAGS = -s -X main.version=$(LOCAL_TAG) -X main.build=$(LOCAL_BUILD) -X main.commit=$(LOCAL_COMMIT)
```
abeaumont: minor: it may be better to use `+=` or `?=` instead of `=`, since having a customizable `LDFLAGS` by the user may be desirable.

Makefile
```diff
@@ -1,39 +1,40 @@
...
+MAKEFILE = Makefile.main
+CI_REPOSITORY = https://github.com/src-d/ci.git
+CI_FOLDER = .ci
+
+# If you need to build more than one dockerfile, you can do so like this:
+# DOCKERFILES = Dockerfile_filename1:repositoryname1 Dockerfile_filename2:repositoryname2 ...
+
+$(MAKEFILE):
+	@git clone --quiet $(CI_REPOSITORY) $(CI_FOLDER); \
+	cp $(CI_FOLDER)/$(MAKEFILE) .;
```
abeaumont: If you just need the Makefile.main file, it may be simpler and more efficient to just get it:

`wget https://raw.githubusercontent.com/src-d/ci/master/Makefile.main`

mcarmonaa: It's how the example does it https://github.com/src-d/ci/blob/master/examples/ci/Makefile

Should I change it anyway?

cli/enry/main.go
mcarmonaa: No reason, let me change it to the no hardcoded version

Makefile
mcarmonaa: Maybe it's better just rename the variable to `LOCAL_LDFLAGS` because in the Makefile.main could be in the future a `LDFLAGS` to build the binaries for releases and it shouldn't be changed

what do you think?

abeaumont: sounds good

.gitignore
mcarmonaa: I have it in my local:

```bash
➜  enry.v1 git:(fix/makefile-ci) ✗ hexdump -c .gitignore
0000000   .   l   i   n   g   u   i   s   t  \n   .   c   i  \n   M   a
0000010   k   e   f   i   l   e   .   m   a   i   n  \n                
000001c
```

but it seems that GH remove the new line

Makefile
abeaumont: hmmm, I'd like to hear what @smola or @erizocosmico have to say then, since they'll know better.

.gitignore
abeaumont: weird, it seems correct now

Makefile
erizocosmico: It could be done with wget, Makefile.main has no other dependencies afaik. Check in with @mcuadros first though and if he approves then also make a PR in the examples.

mcarmonaa:
I opened a PR to discuss about using wget to download only Makefile.main

https://github.com/src-d/ci/pull/37

Makefile
abeaumont: This is being handled here: https://github.com/src-d/ci/pull/37

PR #77, opened 2017-07-14
legal: Replace MIT by Apache2 license

abeaumont:


LICENSE
```diff
@@ -1,19 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "{}"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright {yyyy} {name of copyright owner}
```
bzz: Shall this be replaced with proper copyright?
As per https://github.com/src-d/guide/blob/master/engineering/documents/apache/LICENSE

abeaumont: oh, I didn't know we had those files in the guide. We should certainly use them.
That being said, I don't think the LICENSE file is the place where that replacing should be done since:
* the description doesn't make sense if replaced.
* you're modifying a LICENSE file that should be generic.

abeaumont: Changed

bzz:
Looks great to me

PR #91, opened 2017-08-07
add C exports for enry

erizocosmico:
/cc @mcuadros @smola @mcarmonaa @abeaumont

### DIFF
--- a/shared/enry.go
+++ b/shared/enry.go
@@ -1,0 +1,140 @@
+// +build darwin,cgo linux,cgo
+// +build amd64
+
+package main
+
+import "C"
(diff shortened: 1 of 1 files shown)


shared/enry.go
```diff
@@ -0,0 +1,140 @@
+// +build darwin,cgo linux,cgo
+// +build amd64
```
erizocosmico: perhaps we should not restrict this here, but on the client?

abeaumont: It doesn't seem a big deal, we can leave it here for now and be less restrictive or move to the client in the future if needed.

PR #92, opened 2017-08-08
java bindings for Enry

erizocosmico:
More info in https://github.com/src-d/enry-java/pull/2

Just review last commit, depends on #91 

Can't assign, so: /cc @mcuadros @smola @mcarmonaa @abeaumont @ajnavarro 

Missing:

* [x] CI stuff
* [x] wget JNAerator jar from jitpack 
* [x] update bindings after #93

PR #93, opened 2017-08-08
common: Add filename parameter to GetLanguageByContent function

### LINKED ISSUE #90: GetLanguageByContent does nothing
In `GetLanguageByContent` (https://github.com/src-d/enry/blob/master/common.go#L88) we can see the passed filename is `""`, but in `GetLanguagesByContent` (https://github.com/src-d/enry/blob/master/common.go#L382), which is called by the aforementioned function, returns always nil if the extension is not matched, which always happens because `GetLanguagesByContent` is explicitly passing an empty string.

We should either fix this or remove this exported function, because it does nothing.

abeaumont:
Internal code needs the filename to select a matcher, so not passing
any filename means no language will ever be found otherwise.
Fixes #90.

### DIFF
--- a/common.go
+++ b/common.go
@@ -136,9 +136,9 @@ func GetLanguages(filename string, content []byte) []string {
 	return languages
 }
 
-// GetLanguagesByModeline returns a slice of possible languages for the given content, filename will be ignored.
+// GetLanguagesByModeline returns a slice of possible languages for the given content.
 // It complies with the signature to be a Strategy type.
-func GetLanguagesByModeline(filename string, content []byte, candidates []string) []string {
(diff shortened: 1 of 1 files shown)


erizocosmico:
Ping me when this is merged so I can update #91 and #92

mcarmonaa:
Related to this bug `GetLanguagesByContent`'s comment should be fixed too since it's saying that `filename` will be ignored. Maybe it could be changed in this PR too.

abeaumont:
@mcarmonaa oh, right. I think that's an invariant that is better imposed in code that commented in documentation. I've changed it in all the places I've seen.

abeaumont:
@erizocosmico ping.

PR #96, opened 2017-08-11
Java bindings: add some convenience

bzz:
Minor convenience for JVM clients:
 - add static `Enry.unknownLanguage` 
 - generate `Guess.toString/.equals/.hashCode`

\cc @erizocosmico @abeaumont for review

### DIFF
--- a/java/src/main/java/tech/sourced/enry/Enry.java
+++ b/java/src/main/java/tech/sourced/enry/Enry.java
@@ -5,6 +5,8 @@ import tech.sourced.enry.nativelib.*;
 import static tech.sourced.enry.GoUtils.*;
 
 public class Enry {
+    public static final Guess unknownLanguage = new Guess("", false);
+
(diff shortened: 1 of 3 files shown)


java/src/main/java/tech/sourced/enry/Enry.java
```diff
@@ -5,6 +5,8 @@
 import static tech.sourced.enry.GoUtils.*;
 
 public class Enry {
+    public static final Guess unknownLanguage = new Guess("UNK", false);
```
abeaumont: We use empty string for unknown language in Go, not sure if it'd be fit well here. If so, I'd go with the same convention (could be `null` in Java)

bzz: Empty string sounds good

bzz:
Changed to adhere Go conventions. History re-written in 0a6ed07

abeaumont:
thanks @bzz !

PR #105, opened 2017-09-20
Make jitpack aware of JVM build

bzz:
Until it's published on Maven Central - it would re really nice to be able to use jitpack.com

As java build is in `./java` dir which is not a standard project layout, this PR adds temporarily instructions to work that around using https://jitpack.io/docs/BUILDING/#build-customization

### DIFF
--- a/java/Makefile
+++ b/java/Makefile
@@ -49,10 +49,10 @@ darwin-shared:
 	$(MAKE) darwin-shared
 
 test:
-	sbt clean test
+	./sbt clean test
 
 package:
-	sbt clean assembly
+	./sbt clean assembly
 
 clean:
 	rm -rf $(JAR)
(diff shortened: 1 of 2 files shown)


abeaumont:
@bzz This config is for some kind of cli provided by jitpack which should be called from project root path, right?

bzz:
Jitpack builds projects on it's infrastructure and hosts artefacts in internal maven repository.

It relies on project conventions to find a build system. In our case we want to instruct it to find sbt build not in the root, but rather in `./java`

I think it needs to include `make` as well, to build native dependencies first! Let me update it

abeaumont:
@bzz and jitpack is able to find the yaml config file in a subdir?

bzz:
Yes, thank you! Let me fix that ASAP

bzz:
@abeaumont that should work now

bzz:
Idea was good, but Jitpack [does not have go installed](https://jitpack.io/com/github/bzz/enry/build~jvm-on-jitpack-v1.2.0-ga212202-136/build.log) so this would be not that simple to get it working

PR #106, opened 2017-09-22
Add failing test for empty file content

bzz:
`getLanguage` fails if the content of the file is empty.


bzz:
\cc @abeaumont for review

PR #107, opened 2017-09-22
Java: build using sbt wrapper

bzz:
This allows for `enry-java` build on machine, without assumption of having sbt installed


abeaumont:
Thanks!

PR #113, opened 2017-09-28
Use rubex for faster regular expressions

vmarkovtsev:
I had to rewrite the non-greedy modifier because it is not supported by oniguruma

Alarma! It requires the oniguruma library installed, see https://github.com/moovweb/rubex

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,16 +4,26 @@ go:
   - 1.8
   - tip
 
+addons:
+  apt:
+    packages:
+      - libonig-dev
+
 matrix:
   allow_failures:
     - go: tip
   fast_finish: true
+
+env:
+  - ONIGUMURA=1
+  - ONIGUMURA=0
 
 install:
   - rm -rf $GOPATH/src/gopkg.in/src-d
   - mkdir -p $GOPATH/src/gopkg.in/src-d
   - ln -s $PWD $GOPATH/src/gopkg.in/src-d/enry.v1
   - cd $GOPATH/src/gopkg.in/src-d/enry.v1
+  - if [ "$ONIGUMURA" == "1" ]; then make onigumura; fi
   - go get -v -t ./...
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ matrix:
   fast_finish: true
 
 env:
-  - ONIGUMURA=1
   - ONIGUMURA=0
+  - ONIGUMURA=1
 
 install:
   - rm -rf $GOPATH/src/gopkg.in/src-d
@@ -46,6 +46,8 @@ deploy:
     tags: true
 
 jobs:
+  env:
+    - ONIGUMURA=0    
   include:
     - stage: test
       language: scala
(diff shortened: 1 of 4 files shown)


vmarkovtsev:
This does not pass the test suite. Some of the files are detected wrong.

This gives 2x speedup on the benchmark dataset and 1.6x speedup on the Linux kernel.

Máximo says that we could add an option in Makefile to build with oniguruma; it would rewrite the imports and several regexps which are incompatible.

Currently, I focus on passign the tests.

vmarkovtsev:
I passed TestEnryTestSuite but am currently failing TestGeneratorTestSuite

vmarkovtsev:
I passed the whole test suite. Now making the proper Travis config.

mcuadros:
I found this interesting but always being a optional thing, and not the default, avoiding the requirement of cgo by default.

vmarkovtsev:
@mcuadros This is not enabled by default, the import rewriting that you suggested is done with

```
make onigumura
```

and can be canceled with

```
make revert-onigumura
```

What's left in this PR is adding the documentation.

vmarkovtsev:
All is done on my side, ready to the review.

README.md
```diff
@@ -19,6 +19,31 @@ To build enry's CLI you must run
...
+sudo apt install libonig-dev
+```
+
+To build enry with Onigumura regexps, patch the imports with
+
+```
+make onigumura
+```
+
+nnd then rebuild the project.
```
bzz: and

README.md
```diff
@@ -19,6 +19,31 @@ To build enry's CLI you must run
 it generates a binary in the project's root directory called `enry`. You can move this binary to anywhere in your `PATH`.
 
 
+### Faster regexp engine
+
+[Onigumura](https://github.com/kkos/oniguruma) is CRuby's regular expression engine.
```
bzz: Was confused reading this first time and had the same question as https://github.com/src-d/enry/pull/113#issuecomment-333099095

What do you think about highlighting here an optional nature of this improvement? 
Like`Optionally, you can enable [Origumara] - a CRuby's ..`

vmarkovtsev:
Thanks @bzz fixed

Makefile
```diff
@@ -58,6 +64,15 @@ benchmarks-slow: $(LINGUST_PATH)
 	mkdir -p benchmarks/output && go test -run=NONE -bench=. -slow -benchtime=100ms -timeout=100h >benchmarks/output/enry_samples.bench && \
 	benchmarks/linguist-samples.rb 5 >benchmarks/output/linguist_samples.bench
 
+$(RUBEX_ORIG): %.orig : %
+	sed -i.orig -e 's/"regexp"/regexp "github.com\/moovweb\/rubex"/g' $<
```
vmarkovtsev: This is the universal sed syntax which works under macOS

.travis.yml
```diff
@@ -4,16 +4,26 @@ go:
...
+  apt:
+    packages:
+      - libonig-dev
+
 matrix:
   allow_failures:
     - go: tip
   fast_finish: true
 
+env:
```
abeaumont: Please rebase to master since with the publication of `enry-java`, it has changed quite a lot.

vmarkovtsev:
@abeaumont Done, automatic rebase worked without conflicts, tests pass.

JDK jobs take ONIGUMURA=1, I don't think it's desired, right?

abeaumont:
@vmarkovtsev right, ONIGUMURA should be 0 not only in JDK jobs but also in those jobs inside `jobs` section, that publish the shared libs, to be used by `enry-java`.

vmarkovtsev:
fixed

PR #114, opened 2017-09-28
Add the external test linguist dir from env var

vmarkovtsev:
This allows to use a cached directory with linguist instead of cloning and speeds up the tests by 3x on my local machine.

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -20,6 +20,7 @@ type EnryTestSuite struct {
 	suite.Suite
 	repoLinguist string
 	samplesDir   string
+	cloned       bool
 }
 
 func TestEnryTestSuite(t *testing.T) {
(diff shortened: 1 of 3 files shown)


common_test.go
```diff
@@ -28,22 +29,28 @@ func TestEnryTestSuite(t *testing.T) {
 
 func (s *EnryTestSuite) SetupSuite() {
 	var err error
-	s.repoLinguist, err = ioutil.TempDir("", "linguist-")
-	assert.NoError(s.T(), err)
+	s.repoLinguist = os.Getenv("ENRY_TEST_REPO")
+	s.cloned = len(s.repoLinguist) == 0
```
abeaumont: It may be better to compare strings than lengths: `s.repoLinguist == ""`

internal/code-generator/generator/generator_test.go
```diff
@@ -88,20 +89,26 @@ func TestGeneratorTestSuite(t *testing.T) {
 
 func (s *GeneratorTestSuite) SetupSuite() {
 	var err error
-	s.tmpLinguist, err = ioutil.TempDir("", "linguist-")
-	assert.NoError(s.T(), err)
+	s.tmpLinguist = os.Getenv("ENRY_TEST_REPO")
+	s.cloned = len(s.tmpLinguist) == 0
```
abeaumont: same as above

vmarkovtsev:
@abeaumont Fixed

internal/code-generator/generator/generator_test.go
```diff
@@ -88,20 +89,26 @@ func TestGeneratorTestSuite(t *testing.T) {
...
+	if s.cloned {
+		s.tmpLinguist, err = ioutil.TempDir("", "linguist-")
+		assert.NoError(s.T(), err)
+	}
 
-	cmd := exec.Command("git", "clone", lingustURL, s.tmpLinguist)
-	err = cmd.Run()
-	assert.NoError(s.T(), err)
+	if s.cloned {
```
abeaumont: This second `if` is not needed, it should be inside the previous one

vmarkovtsev:
@abeaumont Fixed

PR #124, opened 2017-10-16
Implement Docsrv

dpordomingo:


Makefile
```diff
@@ -6,16 +6,23 @@ COMMANDS = cli/enry
...
-# DOCKERFILES = Dockerfile_filename1:repositoryname1 Dockerfile_filename2:repositoryname2 ...
-
 $(MAKEFILE):
 	@git clone --quiet $(CI_REPOSITORY) $(CI_FOLDER); \
 	cp $(CI_FOLDER)/$(MAKEFILE) .;
-
 -include $(MAKEFILE)
 
+# Docsrv: configure the languages whose api-doc can be auto generated
+LANGUAGES = go
```
abeaumont: There are also java bindings, should it be reflected here?

Makefile
```diff
@@ -6,16 +6,23 @@ COMMANDS = cli/enry
...
+# Docsrv: configure the directory containing the python sources
+PYTHON_MAIN_DIR ?= ./python
```
abeaumont: There's no python dir

README.md
```diff
@@ -120,7 +120,7 @@ Note that even if enry's CLI is compatible with linguist's, its main point is th
 Java bindings
 ------------
 
-Generated Java binidings using a C shared library + JNI are located under [`java`](java)
+Generated Java binidings using a C shared library + JNI are located under [`java`](https://github.com/src-d/enry/blob/master/java)
```
abeaumont: `docsrv` cannot infer relative references? Shouldn't it be better to add support to infer them instead of forcing every repo to avoid them? (which will be hard)

Makefile
dpordomingo: Java api was not defined as a requirement; it could be added as a feature request.

Makefile
dpordomingo: It's a default value, used if there is python code (that do not interfere if there is no python code)
Anyway, I can delete it from enry for sure (I'll do so in a new PR)

README.md
dpordomingo: It could be proposed as a feature request, but for this first release docsrv only converts `md` to `html` instead of parsing/replacing links

dpordomingo:
@abeaumont `docsrv` would serve a small group of repos in its first version. For that group, I'd PR the needed changes, and after that, I'll document the final interface. We decided to proceed that way to try covering a representative group of projects before deciding a final interface.
Regarding to the necessity of adding a `make` rule for implementing docsrv, it was defined as an initial requirement. In the future, there will be more info in the repository that will be used for docsrv, like logos, descriptions...

dpordomingo:
@abeaumont answered your doubts inline.
Added a commit https://github.com/src-d/enry/pull/124/commits/54de51acf74be953f677d8796bcba117d57c903e with your python suggestion.

PR #127, opened 2017-10-18
make IsDotFile do not treat '.' as true

setekhid:
IsDotFile will treat the current directory '.' as a dot file. This PR will fix it theoretically.

`make test` costs too long, and broken with timeout some times, so those code haven't been passed tests yet.

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -40,7 +40,8 @@ func IsConfiguration(path string) bool {
 
 // IsDotFile returns whether or not path has dot as a prefix.
 func IsDotFile(path string) bool {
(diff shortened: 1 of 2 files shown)


abeaumont:
@setekhid Thanks for the fix! Could you please sign-off the commits to pass the check and comply with the [DCO](https://github.com/src-d/enry/blob/master/DCO)? (It was missing in this repo, sorry).

setekhid:
Done

abeaumont:
Great, thanks again!

Issue #129, opened 2017-11-08
Slice out of range error


ajnavarro:
Calling to `GetLanguage` method, sometimes we receive this error:

panic: runtime error: slice bounds out of range
goroutine 17 [running, locked to thread]:
bytes.Count(0x7f627818c0d0, 0x4b, 0x0, 0x1c420038c20, 0x1, 0x20, 0x0)
    /home/travis/.gimme/versions/go1.8.linux.amd64/src/bytes/bytes.go:62 +0x21d
gopkg.in/src-d/enry%2ev1.getHeaderAndFooter(0x7f627818c0d0, 0x4b, 0x0, 0x0, 0x0, 0x0)
    /home/travis/gopath/src/gopkg.in/src-d/enry.v1/common.go:161 +0x9d
gopkg.in/src-d/enry%2ev1.GetLanguagesByModeline(0x7f627818f4a0, 0x0, 0x7f627818c0d0, 0x4b, 0x0, 0x7f62482e6c08, 0x0, 0x0, 0x0, 0x0, ...)
    /home/travis/gopath/src/gopkg.in/src-d/enry.v1/common.go:142 +0x5c
gopkg.in/src-d/enry%2ev1.GetLanguages(0x7f627818f4a0, 0x0, 0x7f627818c0d0, 0x4b, 0x0, 0x0, 0x0, 0x0)
    /home/travis/gopath/src/gopkg.in/src-d/enry.v1/common.go:126 +0x129
gopkg.in/src-d/enry%2ev1.GetLanguage(0x7f627818f4a0, 0x0, 0x7f627818c0d0, 0x4b, 0x0, 0x1c420038e48, 0x1c42001a500)
    /home/travis/gopath/src/gopkg.in/src-d/enry.v1/common.go:38 +0x55
main.GetLanguage(0x7f627818f4a0, 0x0, 0x7f627818c0d0, 0x4b, 0x0, 0x0, 0x0)
    /home/travis/build/src-d/enry/shared/enry.go:11 +0x55
main._cgoexpwrap_f7db11756761_GetLanguage(0x7f627818f4a0, 0x0, 0x7f627818c0d0, 0x4b, 0x0, 0x0, 0x0)
    command-line-arguments/_obj/_cgo_gotypes.go:58 +0x9a

mcarmonaa:
@ajnavarro how can I reproduce it?

ajnavarro:
I would try to send to `GetLanguagesByModeline` an empty or nil `content` param.

[...]

PR #130, opened 2017-11-08
Fixing getHeaderAndFooter issues

mcarmonaa:
***This PR is related to #129 , please see comments there***

Changes in this PR:
- `data/` directory has been synchronized against `linguist` data
- Fixed failing test `TestGetLanguagesByShebang`
- Added cases with empty or nil conten to `TestGetLanguagesByModeline`
- Now nil or empty content is checked in `getHeaderAndFooter` function though it should be needed, see #129 
- Change go version in travis from `1.8` to `1.8.x` to use the latest minor release of the version. Not sure it this could fix some of our problems. See #129

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.8
+  - 1.9.x
   - tip
 
 addons:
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.8
+  - 1.8.x
```
abeaumont: You could directly bump to 1.9.x

Issue #110, opened 2017-09-28
The results differ from linguist much


vmarkovtsev:
```
sourced@sourced-MacBookPro:/tmp$ git clone https://github.com/src-d/enry &>/dev/null
sourced@sourced-MacBookPro:/tmp$ cd enry
sourced@sourced-MacBookPro:/tmp/enry$ time linguist
99.25%  Go
0.35%   Shell
0.24%   Java
0.08%   Ruby
0.05%   Makefile
0.01%   Scala
0.01%   Gnuplot

real    0m1.945s
user    0m1.848s
sys    0m0.056s
sourced@sourced-MacBookPro:/tmp/enry$ time /home/sourced/Projects/ast2vec/enry 
3.28%    Makefile
63.93%    Go
9.84%    CSV
6.56%    Shell
1.64%    Gnuplot
1.64%    Text
3.28%    Ruby
3.28%    Scala
6.56%    Java

real    0m0.084s
user    0m0.072s
sys    0m0.008s
```

abeaumont:
Thanks for reporting @vmarkovtsev.
`linguist` uses number of bytes to get these stats, while `enry` uses number of files.
We could change it at some point to make it compatible.

vmarkovtsev:
Aaa, this explains a lot! Thanks.

abeaumont:
@mcuadros should we work on this compatibility mode, what do you think?

bzz:
Opt-in compatibility mode would make a lot of sense.

vmarkovtsev:
+1 this. IMHO there should be three modes: number of files, number of lines and number of bytes. All three metrics are used in the scientific papers.

zkry:
Hello, I'd like to work on this 🙂. I created a PR with two flags to improve compatibility with linguist.

The first one is `mode` where you can select byte, file or line and have the files be counted accordingly.

I also noticed that by default linguist excludes non-programming languages so I created a simple flag that does this.

Unfortunately the two commands are not 100% the same as linguist seems to ignore some files. For example on the github.com/istio/istio project, enry counts this file  https://raw.githubusercontent.com/istio/istio/master/addons/servicegraph/js/viz/viz-lite.js while linguist skips it making their results differ a lot.

vmarkovtsev:
That's very cool @zkry 

[...]

PR #139, opened 2018-02-23
Add counting mode, and only programming languages flags. (#110)

zkry:
#### What the commit does
This commit addresses some of the problems in issue #110 to address compatibility with linguist. To do this two main flags were added: one to choose the way the files are counted, and another to exclude the program from counting non-programming languages. When working on the line-counting mode, I noticed a bug in the way non-blank lines were counted where two or more blank lines in a row would throw the count off. It has a test too.

#### Things to look out for
Let me know if there is anything I can improve on the documentation of these flags. Also, do you have any recommendations how I can get the two programs to analize the same files. linguist seems to skip over certain files and I'm not sure why.

### DIFF
--- a/cmd/enry/main.go
+++ b/cmd/enry/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
 	"flag"
@@ -27,6 +28,9 @@ func main() {
 	breakdownFlag := flag.Bool("breakdown", false, "")
 	jsonFlag := flag.Bool("json", false, "")
 	showVersion := flag.Bool("version", false, "Show the enry version information")
+	onlyProg := flag.Bool("prog", false, "Only show programming file types in output")
+	countMode := flag.String("mode", "file", "the method used to count file size. Available options are: file, line and byte")
+
 	flag.Parse()
 
 	if *showVersion {
@@ -45,7 +49,10 @@ func main() {
 	}
 
 	if fileInfo.Mode().IsRegular() {
-		printFileAnalysis(root)
+		err = printFileAnalysis(root)
+		if err != nil {
+			fmt.Println(err)
+		}
 		return
 	}
 
@@ -103,6 +110,11 @@ func main() {
 			}
 		}
 
+		// If we are displaying only prog. and language is not prog. skip it.
+		if *onlyProg && enry.GetLanguageType(language) != enry.Programming {
+			return nil
+		}
+
 		out[language] = append(out[language], relativePath)
 		return nil
 	})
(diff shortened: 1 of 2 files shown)


juanjux:
Hi zkry,

Sorry for waiting so much to reply to you, as a poor excuse we've had some maintainership changes on this project so your PR got in review limbo.

It LGTM, but travis is failing, I'll check the problem in a few hours.

cli/enry/main.go
```diff
@@ -197,10 +278,30 @@ func printFileAnalysis(file string) {
 	)
 }
 
-func getLines(file string, content string) (int, int) {
-	totalLines := strings.Count(content, "\n")
-	nonBlank := totalLines - strings.Count(content, "\n\n")
-	return totalLines, nonBlank
+func getLines(content []byte) (total int, nonBlank int) {
```
dennwc: It will be a bit nicer to use `bufio.Scanner` here because it will not require loading the whole file into memory and the code for blank line detection would be simpler:
```golang
line := scanner.Bytes()
line = bytes.TrimSpace(line)
isBlank := len(line) == 0
```

cli/enry/main.go
```diff
@@ -159,21 +167,94 @@ func printJson(out map[string][]string, buff *bytes.Buffer) {
...
+		t += float64(l)
+	}
+	return t, filesErr
+}
+
+func byteCountValues(files []string) (float64, filelistError) {
+	var filesErr filelistError
+	var t float64
+	for _, fName := range files {
+		f, err := os.Open(fName)
```
dennwc: You are leaking file descriptors here. Close file directly after `Stat`, but before the `if` statement.

zkry:
No problems :) I addressed the issues brought up. The only thing I wasn't sure about was avoiding loading the file all at once. For the getLines function, I did the recommended change and it made the function much simpler but since much of the other enry functions accept []byte, I had a hard time figuring out how to avoid loading the file all at once.

dennwc:
@zkry Commit looks valid, and the second bot is okay with it. Squash looks good, but it looks like the rebase was on top of old master - Github still shows the conflict. Please update the master first before rebasing, or fetch and rebase onto `origin/master`.

@mcuadros Can you please check what's wrong with the bot? It's the second valid PR it rejects.

zkry:
Ok, I've got it now. I was confused because I had gopkg.io as my origin remote. Thanks

dennwc:
Looks good now! Bots are also happy, but CI fails:
```
# gopkg.in/src-d/enry.v1/cli/enry
../../../gopkg.in/src-d/enry.v1/cli/enry/main_test.go:30:28: undefined: getLines
```

dennwc:
Thanks again for contribution and sorry for these conflicts :)

zkry:
And thanks for helping me work through this :)

PR #140, opened 2018-03-28
Rename onigumura to oniguruma

ZJvandeWeg:
This change names the dependency like its called. The link to the
package was correct, but all other references were renamed where I could
find time with git grep.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,15 +15,15 @@ matrix:
   fast_finish: true
 
 env:
-  - ONIGUMURA=0
-  - ONIGUMURA=1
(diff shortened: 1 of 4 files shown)


vmarkovtsev:
@ZJvandeWeg Waiting for your sign-off to merge.

ZJvandeWeg:
@vmarkovtsev Added the sign off, just FYI the link in the widget goes to the non existing CONTRIBUTING file 😉

vmarkovtsev:
@smola This is the problem I had too but I don't remember why Alfredo did not fix it a few months ago.

PR #156, opened 2018-08-28
Refactor Oniguruma integration

ajnavarro:
Instead of use a command to change imports before build, using a build tag to generate the correct binary.

This will allow applications to compile enry using oniguruma with less troubles.

Signed-off-by: Antonio Jesus Navarro Perez <antnavper@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,7 @@ install:
   - mkdir -p $GOPATH/src/gopkg.in/src-d
   - ln -s $PWD $GOPATH/src/gopkg.in/src-d/enry.v1
   - cd $GOPATH/src/gopkg.in/src-d/enry.v1
-  - if [ "$ONIGURUMA" == "1" ]; then make oniguruma; fi
-  - go get -v -t ./...
-
+  - if [ "$ONIGURUMA" == "1" ]; then tags="$tags oniguruma"; fi; go get -v -t --tags "$tags" ./...
 script:
   - make test-coverage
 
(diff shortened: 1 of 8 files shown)


regex/standard.go
```diff
@@ -0,0 +1,25 @@
...
+
+func MustCompile(str string) *regexp.Regexp {
+	return regexp.MustCompile(str)
+}
+
+func QuoteMeta(s string) string {
+	return regexp.QuoteMeta(s)
+}
+
+func NewRegexpArray(r ...*regexp.Regexp) []*regexp.Regexp {
```
dennwc: Looking at this code, it might make sense to define an opaque `Regexp` type in this package that will either contain a Go regexp pointer or an oniguruma's regexp pointer. This way external code will be able to define variables of this type.

Type second option is to use Go type aliases to bind either Go or Oni regexp to a type name in this package. I think this might be the most elegant solution.

.travis.yml
```diff
@@ -24,8 +24,7 @@ install:
   - mkdir -p $GOPATH/src/gopkg.in/src-d
   - ln -s $PWD $GOPATH/src/gopkg.in/src-d/enry.v1
   - cd $GOPATH/src/gopkg.in/src-d/enry.v1
-  - if [ "$ONIGURUMA" == "1" ]; then make oniguruma; fi
-  - go get -v -t ./...
+  - if [ "$ONIGURUMA" == "1" ]; then go get -v -t --tags oniguruma ./... ; else go get -v -t ./... ; fi
```
vmarkovtsev: Better:

```
if [ "$ONIGURUMA" == "1" ]; then tags="$tags oniguruma"; fi
go get -v -t --tags "$tags" ./...
```

ajnavarro:
@vmarkovtsev @dennwc done!

ajnavarro:
@dennwc done.

Issue #158, opened 2018-09-20
Upload latest version to maven


mcarmonaa:
Last version available in maven is `1.6.3` but the latest one is `1.6.5` https://search.maven.org/search?q=g:tech.sourced%20AND%20a:enry-java&core=gav

bzz:
Thank you! Seems like may be a bug in release process somewhere 🤔

bzz:
Indeed https://github.com/src-d/enry/blob/master/java/build.sbt#L3 the Java artefact version is 1.6.3

bzz:
CI is slow today, but release build is in progress https://travis-ci.org/src-d/enry/builds/442671176

bzz:
CI seemed to pass, but maven.org is now down 😕 
<img width="863" alt="screen shot 2018-10-17 at 10 26 22 pm" src="https://user-images.githubusercontent.com/5582506/47111622-b2007a00-d25c-11e8-8ef9-3fe10d3777da.png">

bzz:
🎉 https://search.maven.org/artifact/tech.sourced/enry-java/1.6.6/jar

PR #162, opened 2018-10-17
bump java release artefact version

bzz:
Fixes https://github.com/src-d/enry/issues/158


bzz:
After this is merged, we can release v1.6.6 that is going to include all the goodness.

bzz:
Thank you for prompt reviews!

PR #157, opened 2018-08-30
Bump Go to 1.10

dennwc:
Signed-off-by: Denys Smirnov <denys@sourced.tech>

PR #164, opened 2018-10-17
ci: remove go tip and move default to 1.10

bzz:
As [requested](https://src-d.slack.com/archives/C0J8VQU0K/p1537887553000100) to reduce the load on org Travis.

 - drop go tip
 - migrate to `go 1.10.x`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: go
 
 go:
   - 1.9.x
-  - tip
 
 addons:
   apt:
@@ -10,8 +9,6 @@ addons:
       - libonig-dev
 
 matrix:
-  allow_failures:
-    - go: tip
   fast_finish: true
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
-language: go
+sudo: false
+dist: trusty
 
+language: go
 go:
   - '1.10.x'
 
@@ -53,6 +55,7 @@ jobs:
       jdk: oraclejdk8
 
       install:
+        - gimme version
         - eval "$(curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=$GO_VERSION bash)"
         - go version
         - export GOPATH=$HOME/gopath
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - '1.10'
+  - '1.10.x'
 
 addons:
   apt:
@@ -13,7 +13,7 @@ matrix:
 
 env:
   global:
-    - GO_VERSION='1.10'
+    - GO_VERSION='1.10.x'
   matrix:
     - ONIGURUMA=0
     - ONIGURUMA=1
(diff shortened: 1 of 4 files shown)


bzz:
CI fails on `go 1.11`, so bumped only to `1.10` instead.

CI build time x2 faster:  `18 min 2 sec` -> `10 min 23 sec`

bzz:
CI fails similar to #157  🙃 on JNA failing to find `stddef.h`

```
mkdir -p lib && \
	java -jar ./.jnaerator/jnaerator.jar \
		-package tech.sourced.enry.nativelib \
		-library enry \
		./shared/libenry.h \
		-o ./lib \
		-mode StandaloneJar \
		-runtime JNA;
Auto-configuring parser...
Parsing native headers...
/home/travis/build/bzz/enry/java/shared/libenry.h:8:0: error: File not found: stddef.h in . /usr/include /usr/local/include /usr/include/c++/4.8.4 /usr/include/c++/4.8.4/tr1
```

bzz:
Updated for [changes in go 1.10 compiler](https://go-review.googlesource.com/c/go/+/70890) that treat strings in Cgo a bit different (`GoString->_GoString_`, more details in 5147de9 commit msg).

CI is green now, @dennwc @smola it's ready for review now.

.travis.yml
```diff
@@ -1,22 +1,22 @@
 language: go
 
 go:
-  - 1.9.x
-  - tip
+  - '1.10'
```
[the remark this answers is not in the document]
bzz: Addressed in 4344768

bzz:
CI fails as gimme version on travis is very old (1.2 - 2017-07-09) and the `.x` version support was added in [1.5](https://github.com/travis-ci/gimme/blob/b31d2f24b732062e9d88cd99d9beed37dd7d2b0f/CHANGELOG.md#150---2018-05-29) 😕

bzz:
Always opt in for the latest version of Gimme, CI is green now, ready to be merged

java/src/main/java/tech/sourced/enry/GoUtils.java
```diff
@@ -3,13 +3,14 @@
 import com.sun.jna.Memory;
 import com.sun.jna.Pointer;
 import tech.sourced.enry.nativelib.GoSlice;
-import tech.sourced.enry.nativelib.GoString;
+import tech.sourced.enry.nativelib._GoString_;
```
dennwc: Hmm, interesting. I wonder why it was changed.

bzz: The link to the changes is here https://github.com/src-d/enry/pull/164#issuecomment-431672661

Basically, now there 

```
typedef struct { const char *p; intgo n; } _GoString_;
typedef _GoString_ GoString;
```

and the idea is that one can use `_GoString_` to directly access Go string valuse from C  \wo any coping. There are now helpers in C (`_GoStringLen` and `_GoStringPtr` available only in prolog though) that one can use to directly access content of the string.

This Go-Cgo change resulted in a different code  generated by JNA from C->Java, so our helpers were adapted to use new names.

As noted in 5147de9 - there may be another way of adapting for this change

Issue #163, opened 2018-10-17
Java: automate getting version from the tag on CI


bzz:
Right now for java release artifacts we use hard-coded version number.

That is allright but is prone to errors like #158.

To avoid this, we should have an automation that fetches version number from Git tag on CI, as we do for Go in src-d/ci.

This can be done using `com.typesafe.sbt:sbt-git` as in https://github.com/src-d/jgit-spark-connector/blob/master/build.sbt#L11 from https://github.com/src-d/jgit-spark-connector/issues/146

pratik97:
I would like to take this issue. Can you provide some more information like in which files I should make changes?

bzz:
Closing as #165 was merged and https://github.com/src-d/enry/releases/tag/v1.6.7 released

PR #165, opened 2018-10-21
#163 - Added version automation

pratik97:
#163 - Added version automation


bzz:
Thank you for improvement @pratik97! 

You are on the right way, but seems like some some work is needed -  does the `make && make test` works for you on a local machine for Java bindings under `./java`?

CI fails on Java profile https://travis-ci.org/src-d/enry/jobs/444324780

PR #166, opened 2018-10-21
CI: separate stages for tests and deplay

bzz:
Based on #164 and will be rebased once that one is merged.
Only last 2 commits to be reviewed (c1ac7bb, aac225c).

This is refactoring of `.travis.yml` structure, so it now has:
 - simple and explicit structure using [build stages](https://docs.travis-ci.com/user/build-stages#what-are-build-stages)
 - test and deployment Jobs are run at separate stages (in parallel inside single stage)
 - each job has a clear name in UI in English
 - deployment stage is conditionally triggered only on tag

Effect on CI time (measured by last build time):
 - **before** `8 min 4 sec`
   screen shot 2018-10-21 at 9 01 39 pm

 - **after** `6 min 29 sec`
   screen shot 2018-10-21 at 9 01 45 pm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,7 @@ addons:
 env:
   global:
     - GO_VERSION='1.10.x'
-  matrix:
     - ONIGURUMA=0
-    - ONIGURUMA=1
 
 stages:
   - name: test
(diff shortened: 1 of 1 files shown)


bzz:
rebased on latest master

PR #171, opened 2018-10-24
HOTDIX: fix .travis.yml syntax for conditional deployments

bzz:
Was silently missing `on:` section :(

https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on


bzz:
Merging as a hotfix

Issue #149, opened 2018-05-16
Update or remove benchmarks from README


mcuadros:
Since we got a 2x performance in the new version, we should update the README.md, or remove it.

Soundug:
Hi @mcuadros! Is this issue still open?
I would like to work on this.

bzz:
Hey @Soundug - yes, it very much is!

The best approach is to re-run existing benchmark with the latest release of Enry and update the Readme with results.

bzz:
Latest comparison results using existing benchmark posted in #184

PR #175, opened 2018-10-27
solves issue #149

umang20:
removes benchmark from README.


bzz:
Hi @umang20, thank you for the contribution.

Issue #149 
> we should update the README.md,

https://github.com/src-d/enry/issues/149#issuecomment-434155238 has some more details on how ideal fix for this issue should look like.

PR #176, opened 2018-10-27
Update README.md

vyaspranjal33:
added LICENSE badge


bzz:
Thank you for taking care, but source{d} OSS projects follow this [template](https://github.com/src-d/guide/blob/master/engineering/documents/README.tmpl.md) and generally do not use badges for license as Github shows this in UI quite well.

PR #178, opened 2018-10-31
Update LICENSE

vyaspranjal33:
added template text 
 see this https://github.com/src-d/github-reminder/pull/17#issuecomment-432286706

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2017 Sourced Technologies, S.L.
+    Copyright {yyyy} {name of copyright owner}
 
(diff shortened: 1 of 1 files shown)


smola:
@vyaspranjal33 Thank you for taking the time to contribute. However, this is a change we should do in the same direction (now different repos are doing different things), and we'll have a discussion about it here: https://github.com/src-d/guide/pull/305

Once it's clarified, we'll just update everything. So I would suggest you to not do more PRs about LICENSE text, since they might be closed one way or the other.

Also, when you do a PR to our projects, commits must be signed-off. You can read more about that here: https://probot.github.io/apps/dco/

vyaspranjal33:
@smola  i hope you this 'll get merged :sweat_smile:

PR #179, opened 2018-11-07
Include reference to source{d} Engine

campoy:
Signed-off-by: Francesc Campoy <campoy@golang.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,9 @@ To build enry's CLI you must run
 
 this will generate a binary in the project's root directory called `enry`. You can then move this binary to anywhere in your `PATH`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,9 @@ To build enry's CLI you must run
 
 this will generate a binary in the project's root directory called `enry`. You can then move this binary to anywhere in your `PATH`.
 
+This project is now part of [source{d} Engine](https://sourced.tech/engine),
```
bzz: What do you think about

```suggestion
This project is part of [source{d} Engine](https://sourced.tech/engine),
```

?

campoy: I wanted to keep the `now`, because the project was not created to be part of the engine, but rather the other way around.

PR #183, opened 2018-12-26
Maintenance: batch of minor changes

bzz:
Small cleanup and simplification of Makefile, .gitignore, doc and tests

PR #184, opened 2018-12-26
Update benchmarks to latest Enry and Github-Linguist

### LINKED ISSUE #149: Update or remove benchmarks from README

bzz:
Fixes #149 

Compares enry v1.6.7 vs linguist v7.1.3 using existing benchmarks.

TODOs:
 - [x] update instructions on reproducing benchmarks
 - [x] run "fast" benchmarks locally
 - [x] run all benchmarks, post results (takes 15h)
 - [x] document exact Linguist and Enry revisions in benchmark results 

## New benchmark
distribution


creachadair:
This appears to have been pre-merged with #183. Could you please rebase after that lands so I can see what's left of the difference?

bzz:
> Could you please rebase after that lands so I can see what's left of the difference?

Sure, it's just a workflow that we use in case bunch of PRs depend on one another, and that is the reason the exact commit to be reviewed was linked in PR description.

BTW it is also a convention that a `WIP` in the title means that it's still work in progress and is just open for preliminary feedback, rather then final review, so 👍

creachadair:
> Sure, it's just a workflow that we use in case bunch of PRs depend on one another, and that is the reason the exact commit to be reviewed was linked in PR description.

Makes sense! I just saw a review request pop up and figured therefore that it must be ready.

Issue #155, opened 2018-08-27
Sync with github/linguist


smola:
We synchronized with github/linguist in November 2017, an update is long overdue ;)

smola:
Well, we can wait until YAML heuristics are merged, so we don't need to fix the heuristics.rb parser: https://github.com/github/linguist/pull/4087

smola:
YAML heuristics just landed in master.

bzz:
Makes perfect sense and will be happy start looking into this in couple of weeks, if there is no rush.

bzz:
New XML strategy is going to be handled in a separate issue https://github.com/src-d/enry/issues/192

bzz:
As scope has grown, this has been converted to ☂️ and description was updated, linking to subsequent issues.

bzz:
Plan to cut a v1.7.0 release as soon as heuristics.yml update from #189 lands.

Other 3 items will be left for v1.8.0

bzz:
The last item, a detection accuracy drop, seems to be the most important at the moment. 
I will get back to this, after quantifying the impact under the #246

PR #189, opened 2019-01-09
Sync to linguist 7.2.0: heuristics.yml support

bzz:
Fixes part of the #155 - generate heuristics from `heuristics.yml` instead of parsing `heuristics.rb`.

Major code changes include:
 - update code generator for heuristics strategy in  `./internal/code-generator/heuristics.go` to consume `heuristics.yml` instead of  `heuristics.rb` and produce matchable rule tree
 - re-generated all datastructures in `./data/`
 - re-generated gold test results in `./internal/code-generator/test_files/*.gold`

TODOs:
 - [x] update linguist version & re-generate data
 - [x] adapt content heuristics generator for consuming new `heuristics.yml`
 - [x] update doc in README
 - [x] match language Alias resolution to github linguist one
 - [ ] ~fix new Classifier strategy failures~ -> moved to #194 
 - [x] fix new ByExtension strategy failures
 - [x] re-generate `.gold` test fixtures
 - [x] make CI fail on failing tests 2348fd2
 - [x] verify godoc presence for new things (cleanup \w docs everything else #195)


bzz:
Got back from vacation and keep debugging the case of failing Bayesian classifier for content.
Seems like this has to do with difference in how linguist and enry tokenize the content.

bzz:
CI passes, although there [clearly are failing tests](https://travis-ci.org/src-d/enry/jobs/485412468) :/

```
--- FAIL: TestRegexpEdgeCases (0.00s)
	require.go:794: 
			Error Trace:	common_test.go:45
			Error:      	Received unexpected error:
			            	open .linguist/samples/ActionScript/FooBar.as: no such file or directory
			Test:       	TestRegexpEdgeCases
=== RUN   Test_EnryTestSuite
Cloning Linguist repo to '/tmp/linguist-735567248' as ENRY_TEST_REPO was not set
```

internal/code-generator/generator/documentation.go
```diff
@@ -2,13 +2,15 @@ package generator
...
+
+	"gopkg.in/yaml.v2"
 )
 
-// Documentation reads from fileToParse and builds source file from tmplPath. It complies with type File signature.
-func Documentation(fileToParse, samplesDir, outPath, tmplPath, tmplName, commit string) error {
+// Documentation generates filename matchers in Go for documentation.
+// Based on template in tmplPath, using data from fileToParse and
+// result are saved to outFile. It is of generator.File type.
+func Documentation(fileToParse, _, outFile, tmplPath, tmplName, commit string) error {
```
creachadair: Is there some reason we need to keep the `samplesDir` path if the function doesn't use it? Or if there's a bunch of existing use, can we have a TODO pointing to an issue to fix it later?

Separately, this function has enough parameters that we might want to think about reworking it as a parameter struct at some point, but that's probably out of scope for this change.

internal/code-generator/generator/generator.go
```diff
@@ -1,3 +1,5 @@
+// Package generator provides facilities to generate Golang code from data in
```
creachadair:
```suggestion
// Package generator provides facilities to generate Go code from data in
```

creachadair: Could you please be more specific in the comment about which data you mean?

internal/code-generator/generator/generator.go
```diff
@@ -9,7 +11,10 @@ import (
 	"text/template"
 )
 
-// File is the function's type that generate source file from a file to be parsed, linguist's samples dir and a template.
+// File is a common type for all generator functions.
```
creachadair: Ah, I see why `Documentation` has this weird blank now.

internal/code-generator/generator/generator_test.go
```diff
@@ -85,13 +85,25 @@ type GeneratorTestSuite struct {
...
+	generate    File
+	wantOut     string
+}
+
+func Test_GeneratorTestSuite(t *testing.T) {
 	suite.Run(t, new(GeneratorTestSuite))
 }
 
-func (s *GeneratorTestSuite) SetupSuite() {
+func (s *GeneratorTestSuite) cloneLinguistMaybe() {
```
creachadair: [optional nit] I suggest putting the `maybe` first: `maybeCloneLinguist`.

internal/code-generator/generator/generator_test.go
```diff
@@ -244,8 +241,29 @@ func (s *GeneratorTestSuite) TestGenerationFiles() {
 			wantOut:     mimeTypeGold,
 		},
 	}
+}
 
-	for _, test := range tests {
+func (s *GeneratorTestSuite) TearDownSuite() {
+	if s.cloned {
+		err := os.RemoveAll(s.tmpLinguist)
+		assert.NoError(s.T(), err)
```
creachadair: Do we want to fail the test if the cleanup fails? That isn't really actionable for someone debugging the unit.

internal/code-generator/generator/generator_test.go
```diff
@@ -244,8 +241,29 @@ func (s *GeneratorTestSuite) TestGenerationFiles() {
...
+	}
+}
+
+// UpdateGeneratorTestSuiteGold is a Gold results generation automation.
+// It should only be enabled&run manually on every new linuguist verision
```
creachadair:
```suggestion
// It should only be enabled&run manually on every new Linguist version
```

creachadair: Rather than commenting this out, can we put in a flag?

```go
var updateGold = flag.Bool("update_gold", false, "Update golden test files")
…
func (s *GeneratorTestSuite) TestUpdateGeneratorTestSuiteGold() {
   if !*updateGold {
      s.T().Skip()
   }
   …
}
```

or words to that effect?

internal/code-generator/generator/heuristics_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+	"go/format"
+	"testing"
+	"text/template"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestYamlParsing(t *testing.T) {
```
creachadair: YAML https://github.com/golang/go/wiki/CodeReviewComments#initialisms

internal/code-generator/generator/heuristics_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	require.NoError(t, err)
+	assert.NotNil(t, heuristics)
+
+	// extensions
+	require.NotNil(t, heuristics.Disambiguations)
+	assert.Equal(t, 4, len(heuristics.Disambiguations))
+	assert.Equal(t, 2, len(heuristics.Disambiguations[0].Extensions))
+
+	rules := heuristics.Disambiguations[0].Rules
```
creachadair: Please add a comment to the heuristics.yml file to document that the test cares about the number and order of entries.

internal/code-generator/generator/heuristics_test.go
```diff
@@ -0,0 +1,129 @@
...
+	assert.Equal(t, 2, len(rules[1].And))
+	require.NotNil(t, rules[1].And[0].NegativePattern)
+	assert.Equal(t, "np", rules[1].And[0].NegativePattern)
+
+	rules = heuristics.Disambiguations[3].Rules
+	assert.Equal(t, 1, len(rules))
+	assert.Equal(t, "Linux Kernel Module", rules[0].Languages[0])
+	assert.Equal(t, "AMPL", rules[0].Languages[1])
+
+	// for i, rule := range rules {
```
creachadair: Dead?

internal/code-generator/generator/heuristics_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	// named_patterns
+	require.NotNil(t, heuristics.NamedPatterns)
+	assert.Equal(t, 2, len(heuristics.NamedPatterns))
+	assert.Equal(t, 1, len(heuristics.NamedPatterns["fortran"]))
+	assert.Equal(t, 2, len(heuristics.NamedPatterns["cpp"]))
+}
+
+func TestSingleRuleLoading(t *testing.T) {
+	//given
```
creachadair: I'm not sure what the comments here mean. (But they are missing a leading space)

internal/code-generator/generator/heuristics_test.go
```diff
@@ -0,0 +1,129 @@
...
+			},
+		},
+	}
+
+	_, err := loadHeuristics(parsedYaml)
+	require.Error(t, err)
+}
+
+func TestTemplateMatcherVars(t *testing.T) {
+	parsedYaml, err := parseYaml("test_files/heuristics.yml")
```
creachadair:
```suggestion
	parsed, err := parseYaml("test_files/heuristics.yml")
```

internal/code-generator/generator/heuristics_test.go
```diff
@@ -0,0 +1,129 @@
...
+	require.NoError(t, err)
+
+	heuristics, err := loadHeuristics(parsedYaml)
```
creachadair:
```suggestion
	heuristics, err := loadHeuristics(parsed)
```

internal/code-generator/generator/documentation.go
bzz: It's just to preserve the consistency: all generator functions implement `generator.File` (not sure if it's for a good reason, but they do), but this one was changed a bit to explicitly mark arg as not beeing used.

Generally, one thing to keep in mind is that this PR tries not to improve too much in the current state of the code (there are MANY opportunities) but only touches up on some low hanging fruits in very drive-by fashion

internal/code-generator/generator/generator.go
bzz: Would something like `Package generator provides facilities to generate Go code in data package from ...` work?

Otherwise I'm afraid it would be require to explain whole code-generation approach here, which I'm happy to do but outside of this PR.

internal/code-generator/generator/generator_test.go
bzz: Having runtime switch instead of compile-time sounds really good for convenience, thank your for suggestion!
But at any pace, just having this automation is already very useful for maintainer :)

bzz:
@creachadair all initial feedback addressed in 5fbadc8

creachadair:
> 			            	open .linguist/samples/ActionScript/FooBar.as: no such file or directory

Is that really the path? Or was it meant to be `./linguist/…`?

bzz:
> > ```
> >   	            	open .linguist/samples/ActionScript/FooBar.as: no such file or directory
> > ```
> 
> Is that really the path? Or was it meant to be `./linguist/…`?

Indeed, it's a bug that is expected to trigger the CI failure here and will be fixed right after.
Current issue seems to be that CI uses different build profile than I ran locally.

bzz:
Hooray, CI is failing now! 🎉🎉🎉
2348fd2 was very un-pleasant to figure out, HT @dennwc!

benchmark_test.go
```diff
@@ -28,9 +28,6 @@ var (
 )
 
 func TestMain(m *testing.M) {
-	var exitCode int
-	defer os.Exit(exitCode)
```
dennwc: You may also use anonymous function to still have defer:
```golang
defer func(){
  os.Exit(exitCode)
}()
```
But in this case it's worth setting `exitCode = 1` by default. Sounds more complex than current approach, but wanted to mention it just in case :)

data/heuristics.go
```diff
@@ -0,0 +1,97 @@
...
+	Matcher interface {
+		Match(data []byte) bool
+	}
+
+	Languages struct {
+		langs []string
+	}
+
+	Rule interface {
+		GetLanguages() []string
```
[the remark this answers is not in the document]
creachadair: Separately, it's not clear why a `Rule` has `Languages` and nothing else. I mean of course I understand why it is this way operationally, but the semantics are kind of disjoint: We're saying something about the concrete type we know inhabits this interface, rather than the meaning this package cares about.

This may wind up being a larger design question, in which case we can defer it.

data/heuristics.go
```diff
@@ -0,0 +1,97 @@
...
+func (h *Heuristics) matchString(data string) []string {
+	return h.Match([]byte(data))
+}
+
+// GetLanguages returns languages, defined by this data.Rule.
+func (l *Languages) GetLanguages() []string {
+	return l.langs
+}
+
+type OrRule struct {
```
creachadair: Can we drop the `Rule` suffixes from these? If there are conflicts, they probably deserve a package of their own.

internal/code-generator/assets/alias.go.tmpl
```diff
@@ -1,9 +1,34 @@
...
+	return
+}
+
+
+// ConvertToAliasKey converts language name to a key in LanguageByAliasMap.
+// Following
+//  - internal.code-generator.generator.convertToAliasKey()
+//  - GetLanguageByAlias()
+// conventions.
+// It is here to avoid dependency on "generate" and "enry" packages.
```
creachadair: Can we avoid having to export it?

CONTRIBUTING.md
```diff
@@ -0,0 +1,61 @@
+# source{d} Contributing Guidelines
```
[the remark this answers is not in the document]
bzz: Not really, but it's part of the documentation improvements - all repositories are expected to have such guideline.

This is vanilla from https://github.com/src-d/guide/blob/master/engineering/documents/CONTRIBUTING.md

data/heuristics.go
bzz: Those are rules for auto-generated rule-based matching system, so as a reader I found it convenient to have a common suffix to remind about that.

Can drop the suffix ofc, but I'm not sure how/what does it help with.

common_test.go
```diff
@@ -30,25 +58,24 @@ func Test_EnryTestSuite(t *testing.T) {
...
-	s.samplesDir = filepath.Join(s.repoLinguist, "samples")
-
-	if s.cloned {
-		cmd := exec.Command("git", "clone", linguistURL, s.repoLinguist)
+	s.tmpLinguist = os.Getenv(linguistClonedEnvVar)
+	s.needToClone = s.tmpLinguist == ""
+	if s.needToClone {
+		s.tmpLinguist, err = ioutil.TempDir("", "linguist-")
+		require.NoError(s.T(), err)
+		fmt.Printf("Cloning Linguist repo to '%s' as %s was not set\n",
```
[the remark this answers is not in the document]
bzz: I would say at this point it does not matter much - this code and it's variations is copy-pasted all over the tests and needs refactoring =)

data/heuristics.go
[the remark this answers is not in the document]
bzz: > Separately, it's not clear why a Rule has Languages and nothing else

I'm totally open to suggestions on a better names for these. 
But every rule, if it matches, can produce a number of languages thus the names in here.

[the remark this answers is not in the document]
bzz: I think the source of confusion here is twofold:
 - removing `Get` prefix from `GetLanguages` will not allow the compiler to distinguish between `Languages` type and `Languages()` method. This will result in a panic at a runtime.
 - As Go does not allow all Rules to inherit a data member `langs []string`, it's a trick that I use here: type `Languages` does implement abstract interface `Rule`, and then every concrete `*Rule` embeds the `Languages` type and thus:
     * implements the Rule interface
     * gets access to a data member though `GetLanguages()`

Please let me know if there is a better way of modeling this, but so far this one seems the most concise to me given the language constraints.

I belive this all might also be resolved just by picking a better names for each of those 3 entities mentioned above.

bzz:
@creachadair feedback addressed in c57bc4a and c4f3dbe

CONTRIBUTING.md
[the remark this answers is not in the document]
creachadair: Makes sense.

data/heuristics.go
creachadair: Ah, I didn't realize this was machine generated. Can you make the generator insert a comment at the top of the file along the lines of `// This file was generated, do not edit.` ?

creachadair: As far as the names: It's generally worthwhile to avoid redundant information in names. If all or many of the names in a given package share a common prefix or suffix, that is often a sign that either (a) it should be implicit, or (b) there is a sub-package that can be pulled out.

It's definitely not a blocker by any means, but it does affect the API of the code in use: "AndRule" and "OrRule" are kind of unnatural for someone writing an expression as opposed to simply "And" and "Or". Perhaps if this were package `rule`, we could have `rule.And` and `rule.Or` to get the same effect without decorating the names.

That said: This is not urgent, and even if we decide to do so it can wait.

data/heuristics.go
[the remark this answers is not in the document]
creachadair: There shouldn't be any conflict between the method name and a type name, so even if it were simply `type Rule interface { Languages() []string }` that shouldn't cause problems I think.

That said: My main confusion I think is about names rather than basic structure. If we do not need this interface except to type-assert variant implementations, one simple thing we can do is unexport the interface so it's not part of the public API for the package. In fact, if it's only used in that one place, you can even drop the name:

```go
langs := matcher.(interface{ Languages() []string }).Languages()

// or alternatively
langs := matcher.(interface {
	Languages() []string
}).Languages()
```

…if you want. Though some folks find that harder to read, so YMMV.

Unexporting `rule` would probably assuage most of my concern, which is about clutter in the public API.

data/heuristics.go
bzz: > Ah, I didn't realize this was machine generated. Can you make the generator insert a comment at the top of the file along the lines of `// This file was generated, do not edit.` ?

sorry for confusion - this code is human-written, but is only used inside machine-generated datastructures in `content.go`

bzz:
Thank you for kind explanations and rising the concerns about public API structure, @creachadair !
Will address those first thing tomorrow and let you know.

README.md
```diff
@@ -154,14 +154,15 @@ Generated Java bindings using a C-shared library and JNI are located under [`jav
 Development
 ------------
 
-*enry* re-uses parts of original [linguist](https://github.com/github/linguist) to generate internal data structures. In order to update to the latest upstream and generate the necessary code you must run:
+*enry* re-uses parts of original [linguist](https://github.com/github/linguist) to generate internal data structures. In order to update to the latest upstream and generate all the necessary code you must run:
 
+    git clone https://github.com/github/linguist.git .linguist
```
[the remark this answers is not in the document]
bzz: Moved to https://github.com/src-d/enry/issues/201

data/heuristics.go
bzz: addressed in f299f2050994c8a0be3f197f53707196e761eda6

data/heuristics.go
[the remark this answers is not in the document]
bzz: Thank you for suggestions and explicitly sharing main concern!

I belive it has been addressed in 97ab29a

bzz:
All feedback addressed, @creachadair it's ready for another round 🙏

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+// Matcher checks if the data matches (number of) pattern.
+// Every rule below implements this interface: a rule is matcher that identifies
+// given programming language(s) in case of the match.
+type Matcher interface {
+	Match(data []byte) bool
+}
+
+// languages struct incapsulate data common to every Matcher: all languages
+// that it identifies.
+type languages struct {
```
creachadair: Since this type is returned by the exported members of this package, we should probably export it as well. Otherwise it can be annoyingly inconvenient to use.

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+	return l.langs
+}
+
+// MatchingLanguages is a helper to create new languages.
+func MatchingLanguages(langs ...string) *languages {
+	return &languages{langs}
+}
+
+// Implements a Heuristic.
+type or struct {
```
creachadair: Consider perhaps `orMatcher`, `andMatcher` etc. for these unexported  types? Or make the constructors self-documenting by having them return `Matcher` instead of their concrete type?

    func Or(l Languages, r *regexp.Regexp) Matcher { …

data/rule/rule.go
creachadair: (It's not a hard-and-fast rule, but generally a good idea for exported functions not to return unexported types)

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+	*languages
+	Pattern *regexp.Regexp
```
creachadair: Why exported?

data/rule/rule.go
creachadair: I think you could eliminate the dependency on the `regexp` package entirely from this package: A `*regexp.Regexp` satisfies `rule.Matcher` anyway, at which point the `pattern` field could be a `Matcher` and this argument could as well.

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+	return allDontMatch
+}
+
+// Implements a Heuristic.
+type always struct {
+	*languages
+}
+
+// Always rule always matches. Often is used as a default fallback.
+func Always(l *languages) *always {
```
creachadair:
```suggestion
func Always(*languages) *always {
```

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// languages struct incapsulate data common to every Matcher: all languages
+// that it identifies.
+type languages struct {
+	langs []string
+}
+
+// Languages returns all languages, identified by this Matcher.
+func (l *languages) Languages() []string {
```
creachadair: Need this be passed by pointer? I suggest passing by value so there is no extra indirection required—the slice inside is already a reference value.

creachadair: (then there is no possibility of a nil pointer indirection from improper use, either—the compiler won't allow it)

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Not rule matches if none of the patterns match.
+func Not(l *languages, r ...*regexp.Regexp) *not {
+	return &not{l, r}
+}
+
+// Match implements data.Matcher.
+func (r *not) Match(data []byte) bool {
+	allDontMatch := true
```
creachadair: Simpler, I think:

```go
for _, p := range r.patterns {
  if p.Match(data) {
    return false
  }
}
return true
```

data/rule/rule.go
```diff
@@ -0,0 +1,117 @@
...
+		if p.Match(data) {
+			allDontMatch = false
+			break
+		}
+	}
+	return allDontMatch
+}
+
+// Implements a Heuristic.
+type always struct {
```
creachadair: I don't think this needs a value at all:

```go
type always struct{}

func Always(Languages) Matcher { return always{} }

func (always) Match([]byte) bool { return true }
```

or words to that effect.

creachadair: In any case it doesn't need to be passed by pointer.

data/rule/rule.go
bzz: I agree and appreciate the suggestions. Biggest impact on not having this as a pointer would be such change in the generated code

```
                        rule.Not(
-                               nil,
+                               rule.MatchingLanguages(""),
                                regexp.MustCompile(`(?m)/\*`),
                        ),
```

data/rule/rule.go
bzz: Addressed in ec54891 - now they return a proper `Heuristic`

data/rule/rule.go
bzz: addressed in ec54891

data/rule/rule.go
bzz: Not sure if I understand the suggested change here..

data/rule/rule.go
bzz: Adressed in ec54891

data/rule/rule.go
bzz: 👍this stems from some value in having a single return per function in some other languages.
All implementations simplified in ec54891

data/rule/rule.go
bzz: Would prefer keeping a struct for the benefit of all rules having a very similar structure.
Addressed in ec54891

bzz:
@creachadair thank you for your kind and useful feedback, I belive it all has been addressed and is ready for another pass.

bzz:
New [v7.2.0 of linguist](https://github.com/github/linguist/releases/tag/v7.2.0) has been released, it includes a fix to one of the issues that affected us so will bump to that one.

data/rule/rule.go
bzz: Left private as this is designed to be embedded and the only function that returns it is a factory method that is used for constructing a data-structure literal.

Issue #195, opened 2019-01-29
Ensure exported names in Go packages are properly documented


bzz:
Following [bblfsh/bblfshd#211](https://github.com/bblfsh/bblfshd/issues/211) and to simplify usage of enry as a library for the users (e.g #186) we ned to have GoDoc/JavaDoc/DocStrings on all public methods.

PR #198, opened 2019-01-29
Add documentation comments to package tokenizer.

creachadair:
Although this package is internal, it still exports an API and deserves some
comments. Serves in partial satisfaction of #195.

Signed-off-by: M. J. Fromberger <michael.j.fromberger@gmail.com>

### DIFF
--- a/internal/tokenizer/tokenize.go
+++ b/internal/tokenizer/tokenize.go
@@ -1,4 +1,4 @@
-// Package tokenizer implements file tokenization used by the enry file
+// Package tokenizer implements file tokenization used by the enry content
 // classifier. This package is an implementation detail of enry and should not
 // be imported by other packages.
 package tokenizer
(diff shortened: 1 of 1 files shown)


internal/tokenizer/tokenize.go
```diff
@@ -8,6 +11,9 @@ import (
 
 const byteLimit = 100000
 
+// Tokenize returns classification tokens from content. The tokens returned
```
bzz: I would suggest remove "classification" from the description of this method, as this API is basically exposes only a language-agnostic scanner/lexer that could be used for anything.

internal/tokenizer/tokenize.go
```diff
@@ -1,3 +1,6 @@
+// Package tokenizer implements file tokenization used by the enry file
```
bzz:
```suggestion
// Package tokenizer implements file tokenization used by the enry content
```

bzz:
Let's merge & release & :shipit: :)

bzz:
Merging as CI is green

PR #199, opened 2019-02-04
refactoring: remove un-used code, add go doc

bzz:
Refactoring, consisting of
 - remove unused method `isAuxiliaryLanguage` and `FileCountList`
   in order to reduce public API surfaces (go/java)
 - add GoDoc to public APIs (partially addressing #195)
 - improves CI Java UnitTest job
   It actually uses latest go code now + does not depend on GH org name (by mimicking 
[go_import_path](https://docs.travis-ci.com/user/languages/go/#go-import-path))

TEST PLAN:
 - make test

### DIFF
--- a/enry.go
+++ b/enry.go
@@ -1,3 +1,17 @@
 package enry // import "gopkg.in/src-d/enry.v1"
 
 //go:generate make code-generate
+
+/*
+	Package enry implements multiple strategies for programming language identification.
+
+	Identification is made based on file name and file content using a seriece
+	of strategies to narrow down possible option.
+	Each strategy is available as a separate API call, as well as a main enty point
+
+		GetLanguage(filename string, content []byte) (language string)
+
+	It is a port of the https://github.com/github/linguist from Ruby.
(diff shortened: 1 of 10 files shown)


enry.go
```diff
@@ -1,3 +1,17 @@
 package enry // import "gopkg.in/src-d/enry.v1"
 
 //go:generate make code-generate
+
+/*
```
creachadair: In order for this to count as the package comment, it must immediately precede the `package` clause.

bzz:
Java UnitTest CI fixed in 8ffdf51, PR description updated.

utils.go
```diff
@@ -64,7 +31,8 @@ func IsImage(path string) bool {
 	return false
 }
 
-func GetMimeType(path string, language string) string {
+// GetMIMEType returns a MIME type of a given file based on it's languages.
```
creachadair:
```suggestion
// GetMIMEType returns a MIME type of a given file based on its languages.
```

Issue #191, opened 2019-01-23
enry-java fails on macOS Mojave


smacker:
With dylib from the last release it fails with `Trace/BPT trap: 5`.

```
$ make test
./sbt clean test
[info] Loading project definition from /Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/project
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
[info] Set current project to enry-java (in build file:/Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/)
[success] Total time: 0 s, completed Jan 23, 2019 12:31:29 PM
[info] Updating {file:/Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/}java...
[info] Resolving org.fusesource.jansi#jansi;1.4 ...
[info] Done updating.
[info] Compiling 3 Java sources to /Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/target/classes...
[info] Packaging /Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/target/enry-java-1.6.8-SNAPSHOT.jar ...
[info] Done packaging.
[info] Compiling 1 Java source to /Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/target/test-classes...
[info] Packaging /Users/smacker/go/src/gopkg.in/src-d/enry.v1/java/target/enry-java-1.6.8-SNAPSHOT-tests.jar ...
[info] Done packaging.
make: *** [test] Trace/BPT trap: 5
```

If I build dylib locally on Mojave it works.

creachadair:
Just to verify: By "latest release" you mean https://github.com/src-d/enry/releases/tag/v1.6.7?

creachadair:
Apparently I should finish reading my open threads; this was addressed by #190.

smacker:
@creachadair it's not fixed. I closed the PR because it doesn't work and opened this issue. Please reopen it.

creachadair:
> @creachadair it's not fixed. I closed the PR because it doesn't work and opened this issue. Please reopen it.

Oops! Done, sorry.

bzz:
Thank you guys for logging this in.
Will look into fixing this as soon as I got my hands on machine/VM with Mojave in order to be able to reproduce.

[...]

PR #202, opened 2019-02-20
ci: add go 1.11 

bzz:
Fixes part of the #191 

CI now tests Go test on both, 1.10 and 1.11

Generator tests rely on comparison to the golden gofmt'ed text blobs, but in [go 1.11 gofmt format changed](https://go-review.googlesource.com/c/go/+/122295/) 🤦‍♂️  This change does comparison in tests ignoring Unicode whitespaces (like a `git diff -w`)

`enry-java` is only released using go 1.11

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,24 @@ sudo: false
 dist: trusty
 
 language: go
-go:
-  - '1.10.x'
-
 go_import_path: gopkg.in/src-d/enry.v1
+
+go:
+  - '1.11.x'
+  - '1.10.x'
+env:
+  global:
+    - GO_VERSION_FOR_JVM='1.11.x'
+  matrix:
+    - ONIGURUMA=0
+    - ONIGURUMA=1
+matrix:
+  fast_finish: true
 
 addons:
   apt:
     packages:
       - libonig-dev
-
-env:
-  global:
-    - GO_VERSION='1.10.x'
-    - ONIGURUMA=0
 
 stages:
   - name: test
@@ -24,24 +28,17 @@ stages:
   - name: publish
     if: tag IS present
 
+stage: test
+install:
+  - if [[ -n "$ONIGURUMA" ]]; then tags="$tags oniguruma"; fi; go get -v -t --tags "$tags" ./...
+script:
+  - make test-coverage
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
+
 jobs:
   include:
-    - &golang-unit-tests
-      name: 'golang unitTests'
-      stage: test
-      install:
-        - gimme version
-        - if [ "$ONIGURUMA" == "1" ]; then tags="$tags oniguruma"; fi; go get -v -t --tags "$tags" ./...
-      script:
-        - make test-coverage
-      after_success:
-        - bash <(curl -s https://codecov.io/bash)
-
-    - <<: *golang-unit-tests
-      name: 'golang unitTests, ONIGURUMA=1'
-      env: ONIGURUMA=1
-
-    - name: 'java unitTests'
+    - name: 'java unit-tests'
       stage: test
       language: scala
       jdk: oraclejdk8
@@ -53,8 +50,7 @@ jobs:
         - export TRAVIS_BUILD_DIR=${TRAVIS_HOME}/gopath/src/gopkg.in/src-d/enry.v1
         - cd ${TRAVIS_HOME}/gopath/src/gopkg.in/src-d/enry.v1
       install:
-        - gimme version
-        - eval "$(curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=$GO_VERSION bash)"
+        - eval "$(curl -sL https://raw.githubusercontent.com/travis-ci/gimme/master/gimme | GIMME_GO_VERSION=$GO_VERSION_FOR_JVM bash)"
         - go version
         - echo $PWD; echo $GOPATH
         - go get -v ./...
@@ -67,6 +63,7 @@ jobs:
     - name: 'linux packages'
       stage: release
       install:
+        - go version
         - go get -v -t ./...
       script: make packages
       deploy:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,46 +2,43 @@ sudo: false
...
   - name: publish
     if: tag IS present
 
+stage: test
+install:
+  - if [[ -n "$ONIGURUMA" ]]; then tags="$tags oniguruma"; fi; go get -v -t --tags "$tags" ./...
+script:
+  - make test-coverage
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
creachadair: Should we vendor this script so that a network glitch doesn't cause our CI to stall?

I guess we already do this with Travis, but with gimme we're fetching from Travis's own mirrors. This seems like a third-party operation.

internal/code-generator/generator/generator_test.go
```diff
@@ -283,6 +284,23 @@ func (s *GeneratorTestSuite) TestGenerationFiles() {
...
 		assert.NoError(s.T(), err)
-		assert.EqualValues(s.T(), gold, out, fmt.Sprintf("%v: %v, expected: %v", test.name, string(out), string(gold)))
+
+		expected := withoutSpaces(string(gold))
+		actual := withoutSpaces(string(out))
+		assert.Equal(s.T(), expected, actual, "Test %s", test.name)
+	}
+}
+
+// withoutSpaces is a helper to compare content irregarding of whitespaces.
```
creachadair: I think this could more easily be expressed as

```go
return strings.Join(strings.Fields(str), " ")
```

creachadair:
```suggestion
// withoutSpaces returns a copy of str with whitespace normalized.
```

creachadair: (as it does not do any comparison)

internal/code-generator/generator/generator_test.go
```diff
@@ -283,6 +284,23 @@ func (s *GeneratorTestSuite) TestGenerationFiles() {
...
+// needed as gofmt fomat chnages beteween versions e.g for go 1.10 -> go 1.11
```
creachadair:
```suggestion
// We use this to compare generated source across gofmt changes beteween Go 1.10 and 1.11.
```

.travis.yml
bzz: I think it's their official API for integration https://docs.codecov.io/docs/supported-ci-providers#section-jenkins and all src-d projects use it this way.

Vendoring external resources is a good idea ofc, but I belive better be addressed in a separate issue, treating all resources the same way simultaneously.

internal/code-generator/generator/generator_test.go
bzz: True, but the performance is different:

```
$ go test -bench='BenchmarkWhitespace*' .
goos: darwin
goarch: amd64
pkg: gopkg.in/src-d/enry.v1
BenchmarkWhitespaceJoin-4      	 5000000	       333 ns/op	     192 B/op	       3 allocs/op
BenchmarkWhitespaceBuilder-4   	 5000000	       366 ns/op	      64 B/op	       1 allocs/op
PASS
```

although it's obviously is not most perf critical part, but I would prefer keeping it this way if possible.

bzz:
@creachadair thank you for a prompt review! Feedback addressed in 501fb67 and is ready for another round.

internal/code-generator/generator/drop_whitespace_test.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"strings"
+	"testing"
+)
+
+var wdata = " eityh edsfsdf    dsfds			dsfsdf 	\t\t\t dfsdfsdf sdfsddf"
+
+func dropSpacesJoin(str string) string {
+	return strings.Join(strings.Fields(str), "")
```
creachadair: I recommend normalizing spaces rather than dropping them. Otherwise `"a bc d"` looks equivalent to `"ab cd"` even though they are different token streams.

internal/code-generator/generator/drop_whitespace_test.go
```diff
@@ -0,0 +1,29 @@
+package generator
```
creachadair: Is this file here for your own experimentation? I wouldn't generally advise keeping a benchmark for an internal detail like this, especially one that is only for testing.

internal/code-generator/generator/generator_test.go
creachadair: I'd argue that a difference of 33ns/op on a test helper isn't worthwhile, in contrast to more easily-understood code. But I am not deeply attached to this—I think the more important aspect is to not drop the spaces entirely, because if you do the comparisons could be made incorrect (even though in this case it's unlikely to occur because of the inputs we are using).

FWIW I wrote a token-aware differencing tool for Go source that we could also use, because we ran into this issue with Kythe too—but that might be overkill.

bzz:
All feedback addressed in d9ce949, ready for another pass.

internal/code-generator/generator/generator_test.go
```diff
@@ -283,6 +283,17 @@ func (s *GeneratorTestSuite) TestGenerationFiles() {
...
 		assert.NoError(s.T(), err)
-		assert.EqualValues(s.T(), gold, out, fmt.Sprintf("%v: %v, expected: %v", test.name, string(out), string(gold)))
+
+		expected := normalizeSpaces(string(gold))
+		actual := normalizeSpaces(string(out))
+		assert.Equal(s.T(), expected, actual, "Test %s", test.name)
 	}
 }
+
+// normalizeSpaces returns a copy of str with whitespaces normalizeds.
```
creachadair:
```suggestion
// normalizeSpaces returns a copy of str with whitespaces normalized.
```

internal/code-generator/generator/generator_test.go
```diff
@@ -283,6 +283,17 @@ func (s *GeneratorTestSuite) TestGenerationFiles() {
...
+// We use this to compare generated source as gofmt format may change.
+// E.g for changines beteween Go 1.10 and 1.11 see
```
creachadair:
```suggestion
// E.g for changes beteween Go 1.10 and 1.11 see
```

Issue #204, opened 2019-03-06
Wrong results on django/django


vmarkovtsev:
As @guillemdb found out while doing the ML challenge for source{d}, `enry` reports wrong results for [django/django](https://github.com/django/django). Ground truth:

![image](https://user-images.githubusercontent.com/2793551/53896449-57fa2880-4034-11e9-856d-bf9154d65b61.png)

enry:

![image](https://user-images.githubusercontent.com/2793551/53896458-5d577300-4034-11e9-9293-761fc2629739.png)

bzz:
Thank you for the report, indeed numbers look a bit off! It's documented that some of the strategies do not produce the same results in https://github.com/src-d/enry#divergences-from-linguist 

So far I'm not aware of the attempts to verify the accuracy of enry that goes beyond the `samples` dir of Linguist (as part of the tests, we make sure each strategy agrees with Linguist on every file there) and am looking forward to measuring and improving it under this issue!

Here are the results of  reproducing this comparison locally with `enry` and linguist 7.2.0 CLI

``` 
mkdir /tmp/linguist-django; cd /tmp/linguist-django
git clone --dept 1 git@github.com:django/django.git
cd -

enry -prog /tmp/linguist-django/django
67.85%	Python
31.04%	Modelica
1.02%	JavaScript
0.03%	Makefile
0.03%	Shell
0.03%	Smarty

./bin/github-linguist /tmp/linguist-django/django
95.80%  Python
1.90%   JavaScript
1.66%   HTML
0.63%   CSS
0.01%   Shell
0.00%   Smarty
0.00%   Makefile
```

Good start on this issue would consist of:
 - shell/shell/go script for CI to compare order of magnitude of the results of this breakdown (high-level)
 - dig deeper in CLI code to make sure "apples" are compared to "apples"
 - integration test that over the same codebase file by file and compare results with linguist "gold" results (low-level)

bzz:
Django uses `.mo` files for [localization translation](https://docs.djangoproject.com/en/2.1/topics/i18n/translation/) that are known as [Modelica to linguist](https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L3042) (even if they are not).

So, from a first glance - it seems like in linguist a content classifier was able to differentiate those but on enry it may be the second case of miss-classification from https://github.com/src-d/enry#divergences-from-linguist and most probably due to similar cause as https://github.com/src-d/enry/issues/194 

[...]

PR #205, opened 2019-03-14
Do not return empty lang.

kuba--:
Signed-off-by: kuba-- <kuba@sourced.tech>

So far for `.h` files the `GetLanguages` function return an empty slice (even if internally list of candidates was `C, C++, Objective-C`.
In my opinion it's better to return any potential candidate than nothing.

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -101,9 +101,10 @@ func (s *EnryTestSuite) TestGetLanguage() {
 		expected string
 		safe     bool
 	}{
+		{name: "TestGetLanguage_0", filename: "foo.h", content: []byte{}, expected: "C"},
 		{name: "TestGetLanguage_1", filename: "foo.py", content: []byte{}, expected: "Python"},
 		{name: "TestGetLanguage_2", filename: "foo.m", content: []byte(":- module"), expected: "Mercury"},
-		{name: "TestGetLanguage_3", filename: "foo.m", content: nil, expected: OtherLanguage},
+		{name: "TestGetLanguage_3", filename: "foo.m", content: nil, expected: "MATLAB"},
 		{name: "TestGetLanguage_4", filename: "foo.mo", content: []byte{0xDE, 0x12, 0x04, 0x95, 0x00, 0x00, 0x00, 0x00}, expected: OtherLanguage},
 		{name: "TestGetLanguage_5", filename: "", content: nil, expected: OtherLanguage},
 	}
@@ -276,6 +277,7 @@ func (s *EnryTestSuite) TestGetLanguagesByExtension() {
 		candidates []string
 		expected   []string
 	}{
+		{name: "TestGetLanguagesByExtension_0", filename: "foo.h", expected: []string{"C", "C++", "Objective-C"}},
 		{name: "TestGetLanguagesByExtension_1", filename: "foo.foo", expected: nil},
 		{name: "TestGetLanguagesByExtension_2", filename: "foo.go", expected: []string{"Go"}},
 		{name: "TestGetLanguagesByExtension_3", filename: "foo.go.php", expected: []string{"Hack", "PHP"}},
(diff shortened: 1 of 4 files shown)


kuba--:
If we always return all candidates it will have an impact on performance. So, I assume we start with the _most-likely_ strategies and at the end we check extensions, aliases, etc.
Moreover, if you have a bunch of candidates with the same probability which one would you pick? 
To make the results consistent and testable it will require a stable sort what also will have a bad impact on performance.

vmarkovtsev:
If there are several candidates with the same confidence, I will never trust the library to make any decisions for myself :smile: 

Anyway, I am not proposing to run all the strategies just for the sake of returning a comprehensive list. I am proposing to return all the available information every time, nothing more. That is, if some strategy worked, it had priority over the rest and it output a single item - we return just that item with confidence 1.00.

In other words, it would be better to make the user run `firstLanguage` at their own logic and risk.

kuba--:
So, this is basically what my PR does. I modified `Classifier`, so if you already have some candidates and probability assigned to them, then `Classify` returns a sorted slice. 
Right now, `Classify` expects a content, so even if you have a list of candidates, but no content then you'll get an empty slice.

bzz:
From the first glance, that sounds good. As we do run a number of classifiers that narrow down the options and some(most) of them do not have a notion of confidence we can always do what would be ideal.

Though, as this changes the behavior of the classifier and it's expected output, it will take a bit of time for me verify how this affects clients and accuracy.

If that is not urgent - I will get back to this asap, right after nailing down a bit more pressing issue with accuracy from #204

kuba--:
@bzz - in my opinion it will only change the behaviour for cases where classifier returned an empty slice. It's not a urgent case, but the gitbase issue https://github.com/src-d/gitbase/pull/731
depends on it.

kuba--:
@bzz - this is how I started, but many tests failed. Moreover returning `candidates` if `languages` are empty is unpredictable (every time you'll get the slice in different order), because internally you use a map as a counter and rewrite candidates from a map (again) to the slice, so ordering is not guaranteed. 
If you want to make it consistent you will have to:
- dedup candidates and keep frequency counters somewhere
- sort by counters.
But this is exactly what `classify` does (besides it also reads content and add probability based on read tokens). So, it will duplicate `classify` code in 80% (impact on performance) to return the same result.
Last but not least - most of tools (and for sure tests) rely on `GetLanguages` so if there is any risk that `classify` will break compatibility, then it's the same risk if we return something unexpected from `GetLanguages`

common_test.go
```diff
@@ -339,7 +341,7 @@ func (s *EnryTestSuite) TestGetLanguagesBySpecificClassifier() {
 		{name: "TestGetLanguagesByClassifier_4", filename: filepath.Join(s.samplesDir, "C/blob.c"), candidates: []string{"python", "ruby", "c++"}, classifier: DefaultClassifier, expected: "C++"},
 		{name: "TestGetLanguagesByClassifier_5", filename: filepath.Join(s.samplesDir, "C/blob.c"), candidates: []string{"ruby"}, classifier: DefaultClassifier, expected: "Ruby"},
 		{name: "TestGetLanguagesByClassifier_6", filename: filepath.Join(s.samplesDir, "Python/django-models-base.py"), candidates: []string{"python", "ruby", "c", "c++"}, classifier: DefaultClassifier, expected: "Python"},
-		{name: "TestGetLanguagesByClassifier_7", filename: os.DevNull, candidates: nil, classifier: DefaultClassifier, expected: OtherLanguage},
+		{name: "TestGetLanguagesByClassifier_7", filename: os.DevNull, candidates: nil, classifier: DefaultClassifier, expected: "XML"},
```
bzz: That does not make a lot sense, doesn't it? :)

bzz:
@kuba-- thank you for explanation, both points make sense and my be very well worth mentioning in initial PR description ;) 

I guess what I was trying to say that instead of keep testing for a known bug with unknown impact on accuracy (outside of the small sample) I would prefer a proper fix of what very much looks like a bug in `GetLanguages` :) but that of course could also be done outside of this PR  

https://github.com/src-d/enry/issues/207 was created to track that.

Thanks again!

common_test.go
kuba--: yep, it doesn't. So the question is why classifier for empty list of candidates and empty filename returned a non empty slice.

If you take a look on `classify` code the only difference between what _was_ and what _is_  is the case where the `blob` is nil/empty.
Before we returned an empty slice (even if we had candidates), right now we just don't count extra probability based on tokens. That's it. So the issue with empty candidates and empty file is an another thing. 
We can think about setting some probability threshold (and everything below will get `""`), because I believe in this case XML had very low probability.

Issue #203, opened 2019-02-27
Migrate to src-d/go-oniguruma


bzz:
https://github.com/src-d/go-oniguruma is a maintained fork of the https://github.com/moovweb/rubex
If https://github.com/src-d/go-mysql-server/issues/623 adopts it, Enry should rely on it as well.

bzz:
Fixed by #206

PR #206, opened 2019-03-14
Replace archived rubex by src-d/go-oniguruma

kuba--:
Signed-off-by: kuba-- <kuba@sourced.tech>

### DIFF
--- a/regex/oniguruma.go
+++ b/regex/oniguruma.go
@@ -3,7 +3,7 @@
 package regex
 
 import (
-	"github.com/moovweb/rubex"
+	rubex "github.com/src-d/go-oniguruma"
 )
 
 type EnryRegexp = *rubex.Regexp


go.mod
```diff
@@ -0,0 +1,13 @@
+module gopkg.in/src-d/enry.v1
```
dennwc: I would rather postpone this because we probably want to switch back to `github.com` imports. There is really no point in using `gopkg.in` with modules.

kuba--: Ok. I've just remove module files and left go-oniguruma instead of rubex

bzz:
Looks good to me. I belive it's also worth mentioning in PR description that this addresses #203 

@kuba-- do we expect any performance implications that would make it worth to re-run the benchmark?

kuba--:
@bzz - there is no much changes in library itself. The only reasons why I forked rubex to src-d/go-oniguruma:
- rubex is not in development, anymore (archived) so no prs etc.
- I wanted to suppress compilation warning.
- get rid of all not needed stuff and 2K+ tags

So, technically speaking nothing has been changed in terms of performance, workflow etc.

PR #217, opened 2019-04-04
CLI: always report vendoring for a single file

### LINKED ISSUE #216: enry - linguist single file report logic mismatch
There seems to be a logic mismatch between `enry` and `linguist` on a single file report  - 
`enry` & `linguist` skip vendored files for a `dir` but in case of a single file:
 -  `linguist` returns information [if this file is vendored](https://github.com/github/linguist/blob/master/bin/github-linguist#L73)
 - `enry` doesn't

bzz:
Fixes #216 by always reporting vendoring fact for a single file use-case.


creachadair:
The code coverage is unhappy, but in a trivial way that I do not think matters.

Issue #194, opened 2019-01-28
Bayesian classifier cann't distinguish "SQL" vs "PLpgSQL"


bzz:
Part of the #155.

After update to latest samples in #189, Bayesian classifier test fail to distinguish "SQL" vs "PLpgSQL" based only on content. Classifier weights are different in enry/linguist for the same document https://github.com/src-d/enry/pull/189#issuecomment-457559748

This most probably this has to do with with difference between tokenizations between two projects that going to be addressed in #193

Issue #193, opened 2019-01-28
Port new Tokeniser from Linguist


bzz:
Part of the #155

Right now enry uses content tokenization approach [based on regexps](https://github.com/github/linguist/pull/3846/files#diff-d5179df0b71620e3fac4535cd1368d15L60) from linguist before [v5.3.2](https://github.com/github/linguist/releases/tag/v5.3.2).

This issues is about enry supporting/producing same results as a new, flex-based scanner introduced in [github/linguist#3846](https://github.com/github/linguist/pull/3846).

This is important as it affects Bayesian classifier accuracy and classifier tests in both projects make a strong assumption that all samples can be distinguished by a content classifier alone.

bzz:
Linguist tokenize is defined using flex-based [`tokenizer.l`](https://raw.githubusercontent.com/github/linguist/master/ext/linguist/tokenizer.l).

### 1. flex
Golang does have limited version of it in ported https://gitlab.com/cznic/golex but it is [missing 2 features](https://godoc.org/modernc.org/lex) to in order to be used with the above definition:
 - Trailing context (re1/re2).
 - All flex % prefixed options except %s and %x.

(see logs in details for reproduction instructions)

<details>

```
wget https://raw.githubusercontent.com/github/linguist/master/ext/linguist/tokenizer.l
go get -u modernc.org/golex
golex -o lex.go tokenizer.l

tokenizer.l:35:1: unknown %option "never-interactive yywrap reentrant nounput warn nodefault header-file=\"lex.linguist_yy.h\" extra-type=\"struct tokenizer_extra *\" prefix=\"linguist_yy\""
tokenizer.l:87:16 - "\<[[:alnum:]_!.^/?-]+              {" - trailing context not supported
tokenizer.l:103:15 - "[[:alnum:]_.@#^/*]+                {" - trailing context not supported
```
</details>

At this point it's a hard to estimate the effort of adding those features upstream.

### 2. ragel
Instructive [go-nuts thread on this subject](https://groups.google.com/forum/#!topic/golang-nuts/h5eR60AXrB0) points out worth trying a bit more complex solution, similar to discussion in #167, based on [ragel](http://www.colm.net/open-source/ragel/), another FSM generator that can be "compiled" to Go code. That would only require porting 1 file `.l` -> `.rl` which is much more manageable effort.

bzz:
https://github.com/src-d/enry/issues/193#issuecomment-461954608 updated to include another option of using existing flex-based tokenizer though cgo.

PR #218, opened 2019-04-08
New, optional flex-based tokenizer

bzz:
This change partially address https://github.com/src-d/enry/issues/193

It follows option 3 from https://github.com/src-d/enry/issues/193#issuecomment-461954608

 - a new native tokenizer implementation, hidden behind `-tags flex`
 - existing native code in C from Linguist project, generated by Flex from [tokenizer.l](https://raw.githubusercontent.com/github/linguist/master/ext/linguist/tokenizer.l) is added under `internal/tokenizer/flex/lex.linguist_yy.{h,c}`
 - `tokenizer_c.go` is a manual transliteration to Go of the parser from [linguist.c](https://github.com/github/linguist/blob/master/ext/linguist/linguist.c#L12)

The benchmark before:
```
$ go test -bench=. -run=Benchmark. ./internal/tokenizer
BenchmarkTokenizer_BaselineCopy-4   	 5000000	       304 ns/op	    1536 B/op	       1 allocs/op
BenchmarkTokenizer-4                	    2000	    672592 ns/op	   58205 B/op	     435 allocs/op
```

after
```
$ go test -bench=. -run=Benchmark. -tags flex ./internal/tokenizer
BenchmarkTokenizer_BaselineCopy-4   	 5000000	       300 ns/op	    1536 B/op	       1 allocs/op
BenchmarkTokenizer-4                	   20000	     69302 ns/op	    5064 B/op	     150 allocs/op
```

so it's x10 faster.

But because it's a cgo and it complicates the build I suggest to merge it, keeping behind the tag.

Right now this is only useful for debugging #194 and is ready to be merged without affecting the releases.

But if we agree on it's value for downstream uses for performance reasons, some extra work is going to be needs (in subsequent PR: documentation + updated classifier assets, based on this tokenizer that do not overwrite the default one) before allowing it's usage that goes beyond the internal dev debugging workflow. It is going to be tracked under https://github.com/src-d/enry/issues/193


internal/tokenizer/tokenize_test.go
```diff
@@ -89,27 +90,45 @@ var (
...
 			}
 		})
 	}
 }
+
+func BenchmarkTokenizer_BaselineCopy(b *testing.B) {
+	b.ReportAllocs()
+	for i := 0; i < b.N; i++ {
+		for _, test := range Tests {
+			test.content = append([]byte(nil), test.content...)
```
creachadair: Is there any reason to replace `test.content` with the result here? 

I worry that we're mostly testing the speed of the memory allocator here. If we really care only about copy as a baseline, I suggest pre-allocating:

```go
var size int
for _, test := range Tests {
   if len(test.content) > size {
      size = len(test.content)
   }
}
buf := make([]byte, 0, size)
for _, test := range Tests {
   _ = append(buf[:0], test.content...)
}
```

creachadair: Also note that we are capping Tokenize calls at `byteLimit`, but not doing so for our baseline copies. I think we should probably do the same in both cases for a fair baseline.

internal/tokenizer/flex/tokenize_c.go
```diff
@@ -0,0 +1,73 @@
...
+	var _len C.ulong
+	var r C.int
+
+	_len = C.ulong(len(content))
+	cs := C.CBytes(content)
+	defer C.free(unsafe.Pointer(cs))
+
+	C.linguist_yylex_init_extra(&extra, &scanner)
+	buf = C.linguist_yy_scan_bytes((*C.char)(cs), _len, scanner)
+
```
creachadair:
```suggestion
```

internal/tokenizer/flex/tokenize_c.go
```diff
@@ -0,0 +1,73 @@
...
+// #include "linguist.h"
+// #include "lex.linguist_yy.h"
+// int linguist_yywrap(yyscan_t yyscanner) {
+// 	return 1;
+// }
+import "C"
+import "unsafe"
+
+const maxTokenLen = 32
+
```
creachadair:
```suggestion
```

internal/tokenizer/flex/tokenize_c.go
```diff
@@ -0,0 +1,73 @@
...
+// #include <stdlib.h>
+// #include "linguist.h"
+// #include "lex.linguist_yy.h"
+// int linguist_yywrap(yyscan_t yyscanner) {
+// 	return 1;
+// }
+import "C"
+import "unsafe"
+
+const maxTokenLen = 32
```
creachadair:
```suggestion
const maxTokenLen = 32  // bytes
```

internal/tokenizer/tokenize_c.go
```diff
@@ -0,0 +1,16 @@
+// +build flex
+
+package tokenizer
+
+import "gopkg.in/src-d/enry.v1/internal/tokenizer/flex"
+
+// Tokenize returns language-agnostic lexical tokens from content. The tokens
+// returned should match what the Linguist library returns. At most the first
+// 100KB of content are tokenized.
```
creachadair: In order to avoid skew, maybe we should export the constant:

```go
// ByteLimit defines the maximum prefix of an input text that will be tokenized for classification.
const ByteLimit = 100000
…
// Tokenize returns lexical tokens from content. The tokens returned match what
// the Linguist library returns. At most the first ByteLimit bytes of content are tokenized.
```

YMMV.

Also possibly worthy of note, splitting at a byte offset means we might partition a multibyte unicode character in the middle of a token, but I think that probably doesn't matter in most cases—the only concern is if that results in a false shibboleth for some language, but I think that is probably just theoretical.

internal/tokenizer/tokenize_test.go
bzz: > I worry that we're mostly testing the speed of the memory allocator here

sorry for may be a miss-leading name of the test, but including allocation is expected in this case - it mimics the behaviour of current tokeniser, [up to the point of tokenisation](https://github.com/src-d/enry/blob/ab3c26b46dfaf8281bf78f2e998111e6a40829bf/internal/tokenizer/tokenize.go#L22).

But also it exists just to provide a feeling of an order of magnitude of time that it takes of doing everything, but actual tokenisation, rather than some fine-grained apple-to-apple comparison.

> Is there any reason to replace test.content with the result here?
Not really, will update

bzz:
@creachadair all feedback addressed, ready for another round.

internal/tokenizer/tokenize.go
```diff
@@ -8,12 +8,12 @@ import (
 	"gopkg.in/src-d/enry.v1/regex"
 )
 
-// Tokenize returns language-agnostic lexical tokens from content. The tokens
-// returned should match what the Linguist library returns. At most the first
-// 100KB of content are tokenized.
+// Tokenize returns lexical tokens from content. The tokens returned should match what
```
creachadair: [optional] I suggest making the comment correct-as-intended and separating corrections into a removable paragraph, e.g.,

```
// Tokenize returns lexical tokens from content. The tokens returned match what
// the Linguist library returns. At most the first ByteLimit bytes of content are tokenized.
//
// BUG: Until https://github.com/src-d/enry/issues/193 is resolved, there are some
// differences between this function and the Linguist output.
```

Rationale: This makes it clear what the intent is, and makes it trivial to update the comment when the fix is in: Just delete the BUG paragraph.

(This is not crucial, however—what you wrote is also fine).

internal/tokenizer/tokenize_c.go
```diff
@@ -4,9 +4,10 @@ package tokenizer
 
 import "gopkg.in/src-d/enry.v1/internal/tokenizer/flex"
 
-// Tokenize returns language-agnostic lexical tokens from content. The tokens
-// returned should match what the Linguist library returns. At most the first
-// 100KB of content are tokenized.
+// Tokenize returns lexical tokens from content. The tokens returned match what
```
creachadair: I feel like this comment and the other `Tokenize` comment (in `tokenize.go`) should probably match, up to the specific differences (viz., the known bug in the above).

internal/tokenizer/tokenize.go
bzz: Thank you Michael, that does make a lot of sense and thank you for a concrete suggestions!

internal/tokenizer/flex/tokenize_c.go
```diff
@@ -0,0 +1,71 @@
...
+	var r C.int
+
+	_len = C.ulong(len(content))
+	cs := C.CBytes(content)
+	defer C.free(unsafe.Pointer(cs))
+
+	C.linguist_yylex_init_extra(&extra, &scanner)
+	buf = C.linguist_yy_scan_bytes((*C.char)(cs), _len, scanner)
+
+	ary := []string{}
```
dennwc:
```suggestion
	var ary []string
```
[source](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices)

internal/tokenizer/tokenize_test.go
```diff
@@ -89,27 +90,48 @@ var (
 		"XHTML", "sample", "file", "type", "#example", "background", "color", "yellow", "id", "Just", "a", "simple", "XHTML", "test", "page.",
 		"-", "|", "+", "&&", "<", "<", "-", "!", "!", "!", "=", "=", "!", ":", "=", ":", "=", ",", ",", "=", ">", ">", "=", "=", "=", "=", ">",
 		"'", ",", ">", "=", ">", "=", "=", ">", "=", ">", ":", ">", "=", ">"}
-)
 
-func TestTokenize(t *testing.T) {
-	tests := []struct {
+	Tests = []struct {
```
dennwc: Does it need to be exported?

internal/tokenizer/flex/tokenize_c.go
```diff
@@ -0,0 +1,71 @@
...
+// }
+import "C"
+import "unsafe"
+
+const maxTokenLen = 32 // bytes
+
+// TokenizeFlex implements tokenizer by calling Flex generated code from linguist in C
+// This is a transliteration from C https://github.com/github/linguist/blob/master/ext/linguist/linguist.c#L12
+func TokenizeFlex(content []byte) []string {
+	var buf C.YY_BUFFER_STATE
```
dennwc: Merge into a single `var` block

internal/tokenizer/flex/tokenize_c.go
```diff
@@ -0,0 +1,71 @@
...
+// #include "linguist.h"
+// #include "lex.linguist_yy.h"
+// int linguist_yywrap(yyscan_t yyscanner) {
+// 	return 1;
+// }
+import "C"
+import "unsafe"
+
+const maxTokenLen = 32 // bytes
+
```
dennwc: I would also propose to include a `go:generate` directive that updates the C tokenizer, if it makes sense.

internal/tokenizer/flex/tokenize_c.go
bzz: I know what you mean, but would prefer to keep it as close to original as possible as this is almost [line-by-line transliteration from C](https://github.com/github/linguist/blob/master/ext/linguist/linguist.c#L12)

internal/tokenizer/flex/tokenize_c.go
bzz: I think I did not communicate this well enough, but the only code that is somehow "generated" in this PR is the C `lex.linguist_yy.{h,c}`, that is vendored inside Linguist and we are using exact copy. 

In the long run, if this tokeniser turns out to be useful beyond classifier accuracy debugging - we could add C code generation script from the flex grammar, but I think it's ok to do that after #193

internal/tokenizer/flex/tokenize_c.go
bzz: This would result in returning `nil` instead of empty slice and that would be inconsistent with current impl of tokenizer API, so I think it's better to keep it as is.

internal/tokenizer/tokenize_test.go
bzz: addressed in cbb4c68

bzz:
@dennwc all feedback addressed, ready for another round.

internal/tokenizer/flex/tokenize_c.go
dennwc: Sorry, I haven't meant to regenerate it. `go:generate` may also be used to `wget` the latest source, for example. Not critical, just a thought.

internal/tokenizer/flex/tokenize_c.go
dennwc: The key word is "almost" :) Unless it's machine-generated, I think it makes sense to adhere to Go conventions where possible. It's not a bit change anyway. Still, it's optional, so feel free to ignore.

internal/tokenizer/flex/tokenize_c.go
dennwc: I'm not sure if it matters from the API point of view - it's pretty rare to test for slice nil-ness.

PR #220, opened 2019-04-09
Skip uploading useless jar as part of Github release

### LINKED ISSUE #208: release: stop publishing .jar without native lib
As part of the release process 2 .jar files are published

 1. on Maven e.g [v1.7.2](https://repo1.maven.org/maven2/tech/sourced/enry-java/1.7.2/)
   `enry-java-1.7.2.jar                               2019-03-15 18:07   8078217`

 2. on GH Release e.g [v1.7.2](https://github.com/src-d/enry/releases/tag/v1.7.2)
   `enry-java-1.7.2.jar 4.65 KB`

Second one does not include native library and is useless.

We should either upload same full jar to GH releases, or stop doing that altogether to avoid the confusion.

bzz:
Fixes #208 - in order to avoid possible confusion, it follows [other projects](https://github.com/src-d/siva-java/releases) and skips uploading useless jar on Github release.

Uploading assembly jar instead would require renaming `enry-assembly-<version>.jar` to `enry-<version>.jar` (probably bt reading TravisCI env var to get version number from the tag) in the shell, and although it's a just a several bash LoC, that seems like an unnecessary future maintenance burden anyway as Maven is universally used to distribute release artifacts in JVM world.


bzz:
> But why were we doing this in the first place?

Not sure, my hypothesis is that once it might have been a slim jar that was released so it "just worked", and then we moved to releasing a fat-jar instead, but CI automation was not adjusted.

PR #221, opened 2019-04-09
WIP CI: update to 16.04 + JDK9

### LINKED ISSUE #223: Release: use non-EOLed JDK version
JDK version that we are using for `enry-java` release is oracle JDK 8 which EOLed 1 year ago (and e.g is not available on newer linux 16.04 CI images any more)

This issue is about:
 - [x] update JDK to maintained LTS version e.g openJDK 11
 - [x] update sbt version to 1.x, that works on newer JDK (current 0.13 does not)
 - [x] update sbt shell wrapper and plugin version, that support 1.0
 - [ ] update `build.sbt`, adjusting for plugin changes (fat-jar assembly fails by default)
 - [ ] make sure released version can be run on older JDK (one that Infra uses for the clusters)

Initial work was done under #221 but still a consensus on min JDK version is needed.

bzz:
Fixes #223 

Updates done for CI:
  - [x] switch ubuntu 14.04 -> 16.04
  - [x] bump JDK to a minimal ~still supported~ version of `oraclejdk9`. 
  - [x] update `sbt` wrapper
  - [x] update `sbt` version `0.13.x` -> `1.x`
  - [x] update all `sbt` plugins versions
  - [ ] adapt `build.sbt` to the new plugins apis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ jobs:
     - name: 'java unit-tests'
       stage: test
       language: scala
-      jdk: oraclejdk9
+      jdk: openjdk10
       before_install:
         - export CGO_ENABLED=1
         # mimics exact behavior of 'go_import_path' for non-go build image
@@ -126,7 +126,7 @@ jobs:
     - name: 'java: publish to maven'
       stage: publish
       language: scala
-      jdk: oraclejdk9
+      jdk: openjdk10
       before_install:
         - export CGO_ENABLED=1
         # mimics exact behavior of 'go_import_path' for non-go build image
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: trusty
+dist: xenial
 
 language: go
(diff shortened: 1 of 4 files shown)


bzz:
For anyone interested in this journey though CI update rabbit-hole - check individual commit messages.

The root cause seems to be https://github.com/golang/go/issues/31293 

Going to wait for a new release instead and file a separate issue for updating to newer JDK + sbt versions where this work can be reused.

bzz:
> bradfitz commented 7 hours ago
> we didn't get a new release out today, but we should tomorrow.

bzz:
#223 filed to track further progress of this one.

Needs to be rebased on lates master after #222 lands:
 - drop first 3 commits 
 - rework PR description, as this becomes initial work on #223 instead of a fix

bzz:
Leaving it as-is for now, until there is a consensus on JDK version we want to support in #223

bzz:
Going to close this and just re-use later, as soon as there is progress on #223

PR #222, opened 2019-04-11
CI: workaround for incompatible compiler release binary 

bzz:
Instead of full distro update to get newer compiler https://github.com/src-d/enry/pull/221 - fix last known good go release versions on CI.

This will un-block the next, before-modules release.  

`.x` go version will be restored as soon as both new 1.11 and 1.12 are released https://github.com/src-d/enry/pull/221#issuecomment-481406314

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,11 @@ language: go
 go_import_path: gopkg.in/src-d/enry.v1
 
 go:
-  - '1.11.x'
-  - '1.10.x'
+  - '1.11.6'
+  - '1.12.1'
 env:
   global:
-    - GO_VERSION_FOR_JVM='1.11.x'
+    - GO_VERSION_FOR_JVM='1.11.'
   matrix:
     - ONIGURUMA=0
     - ONIGURUMA=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ go:
   - '1.12.1'
 env:
   global:
-    - GO_VERSION_FOR_JVM='1.11.'
+    - GO_VERSION_FOR_JVM='1.11.1'
+    - CGO_ENABLED=0
   matrix:
     - ONIGURUMA=0
     - ONIGURUMA=1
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,15 +1,15 @@
...
 dist: trusty
 
 language: go
 go_import_path: gopkg.in/src-d/enry.v1
 
 go:
-  - '1.11.x'
-  - '1.10.x'
+  - '1.11.6'
+  - '1.12.1'
```
creachadair: Are we using 1.12.1 instead of 1.12.3 for some specific reason, and if so can you please add a `# brief comment` with a link to the (or an) issue?

internal/code-generator/generator/heuristics.go
```diff
@@ -51,7 +51,7 @@ func loadHeuristics(yaml *Heuristics) (map[string][]*LanguagePattern, error) {
 		// unroll to a single map
 		for _, ext := range disambiguation.Extensions {
 			if _, ok := patterns[ext]; ok {
-				return nil, fmt.Errorf("cannt add extension '%s', it already exists for %q", ext, patterns[ext])
+				return nil, fmt.Errorf("cannt add extension '%s', it already exists for %+v", ext, patterns[ext])
```
creachadair:
```suggestion
				return nil, fmt.Errorf("cannot add extension %q, it already exists for %+v", ext, patterns[ext])
```

.travis.yml
```diff
@@ -122,6 +123,7 @@ jobs:
       language: scala
       jdk: oraclejdk8
       before_install:
+        - CGO_ENABLED=1
```
creachadair: Do you want `export CGO_ENABLED=1` here? I assume you want it in the env, not just in the shell?

.travis.yml
```diff
@@ -30,7 +30,7 @@ stages:
 
 stage: test
 install:
-  - if [[ -n "$ONIGURUMA" ]]; then tags="$tags oniguruma"; fi; go get -v -t --tags "$tags" ./...
+  - if [[ "${ONIGURUMA}" -gt 0 ]]; then export tags="${tags} oniguruma"; CGO_ENABLED=1; fi; go get -v -t -tags "${tags}" ./...
```
creachadair: I _believe_ if you want `CGO_ENABLED` to work you'll either have to export it into the environment or put it on the command-line with the `go …` command, e.g., `CGO_ENABLED=1 go get …`.

.travis.yml
bzz: 👍 yes, ok, will move it from PR description here as well

.travis.yml
bzz: Good catch! Although it works without export as well

bzz:
@creachadair all feedback addressed, ready for another round!

PR #224, opened 2019-04-11
HOTFIX: enable cgo for linux native lib release on CI

bzz:
It was only explicitly enabled for non-linux release build profiles before.


bzz:
[Breaks the release](https://travis-ci.com/src-d/enry/builds/108036010) otherwise.

PR #227, opened 2019-04-17
CI: fix cgo profiles

bzz:
- [x] simple tokeniser test case, to reproduce the difference on RE2 vs oniguruma
 - [x] a fix, by using `MustCompileASCII()` from https://github.com/src-d/go-oniguruma/pull/1

### DIFF
--- a/internal/tokenizer/tokenize_test.go
+++ b/internal/tokenizer/tokenize_test.go
@@ -115,6 +115,13 @@ func TestTokenize(t *testing.T) {
 	}
 }
 
+func TestTokenizerLatin1AsUtf8(t *testing.T) {
+	content := []byte("th\xe5 filling") // `th� filling`
+	t.Logf("%v - %q", content, string(content))
+	tokens := Tokenize(content)
+	require.Equal(t, 3, len(tokens))
+}
+
 func BenchmarkTokenizer_BaselineCopy(b *testing.B) {
 	b.ReportAllocs()
 	for i := 0; i < b.N; i++ {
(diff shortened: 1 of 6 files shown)


bzz:
Waiting for https://github.com/src-d/go-oniguruma/pull/1 to be merged to migrate

internal/tokenizer/tokenize_test.go
```diff
@@ -115,6 +115,45 @@ func TestTokenize(t *testing.T) {
...
+}
+
+func TestRegexpOnInvalidUtf8(t *testing.T) {
+	origContent := []struct {
+		bytes  []byte
+		tokens []string
+	}{
+		{[]byte("th\xe0 filling"), []string{"th", "filling"}},   // `th� filling`
+		{[]byte("th\u0100 filling"), []string{"th", "filling"}}, // `thĀ filling`
+		{[]byte("привет, как дела?"), []string{}},               // empty, no ASCII tokens
```
creachadair: не плохо, спасибо! 😄

internal/tokenizer/tokenize_test.go
```diff
@@ -115,6 +115,45 @@ func TestTokenize(t *testing.T) {
...
+	}
+	require.Equal(t, 3, len(tokens))
+}
+
+func TestRegexpOnInvalidUtf8(t *testing.T) {
+	origContent := []struct {
+		bytes  []byte
+		tokens []string
+	}{
+		{[]byte("th\xe0 filling"), []string{"th", "filling"}},   // `th� filling`
```
creachadair: [optional] This is completely reasonable as you wrote it, but I often find it a little easier to read if the test input is given as a plain `string` and converted to `[]byte` at the point of use, e.g.,

```go
origContent := []struct {
    input string
    tokens []string
}{
   {"th\xe0 filling", []string{"th", "filling"}},
   …
}
…
input := []byte(content.input)
// … the rest of the test.
```

When there are a bunch of test cases, it adds up. It's up to you, though.

bzz: Thank you for suggestion, addressed in f3ceaa6

PR #233, opened 2019-07-19
Implement getting HTML color code for languages

lafriks:
Adds utility function `GetColor(language string) string`


lafriks:
Is there anything else I need to do to get this merged?

creachadair:
> Is there anything else I need to do to get this merged?

Sorry for the delay. @bzz who maintains this repo is out of the office this week, but should be able to take a look when he gets back. The change looks fine from a functional perspective, but I'd like Alex's feedback on whether it makes sense to include it.

internal/code-generator/assets/colors.go.tmpl
```diff
@@ -0,0 +1,7 @@
+package data
+
+var LanguagesColor = map[string]string{
```
bzz: @lafriks here and below, could you also please make sure all exported functions/variables are documented? Thanks!

bzz:
Sorry for delay, great job @lafriks, LGTM.

It would be also interesting to know if you could share about your use-case for this feature!

lafriks:
@bzz no problems, I'm one of core Gitea contributors and currently implementing repository language statistics bar like in Github, see back-reference above for Gitea PR that is currently wip

lafriks:
Btw thanks for for this great library that made this possible to add that to Gitea ;)

PR #234, opened 2019-08-02
Sync to linguist v7.5.1

lafriks:


lafriks:
Tests still fail, will check them

lafriks:
Added special case for `osascript` interpreter in shebang detection if it is specified with `-l` argument

lafriks:
Will update when dependent PR's are merged

bzz:
@lafriks thank you for the fixes, all the PRs are merged now!

After rebasing, would you be so kind to also document a new extension-based heuristic failure on samples for `.rno` at the https://github.com/src-d/enry#divergences-from-linguist?

As soon as this is merged, I'll 🔪 a new release.

lafriks:
Rebased, I don't think there is new failures except for already documented `samples/SQL/drop_stuff.sql`. All other tests now should succeed

lafriks:
ooooh... I finally got what you did mean, it's about unsupported regexp. Added to docs

lafriks:
No problem, all Your requests had valid point, I'm also for clean and traceable code ;)

PR #236, opened 2019-08-05
Improve detection of unsupported regexp syntax

lafriks:
Part 1 from #234

### DIFF
--- a/internal/code-generator/generator/heuristics.go
+++ b/internal/code-generator/generator/heuristics.go
@@ -160,11 +160,12 @@ func parseYaml(file string) (*Heuristics, error) {
 
 // isUnsupportedRegexpSyntax filters regexp syntax that is not supported by RE2.
 // In particular, we stumbled up on usage of next cases:
+// - lookbehind & lookahead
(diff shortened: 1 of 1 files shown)


internal/code-generator/generator/heuristics.go
```diff
@@ -164,7 +164,7 @@ func parseYaml(file string) (*Heuristics, error) {
 // - backreference
 // For referece on supported syntax see https://github.com/google/re2/wiki/Syntax
 func isUnsupportedRegexpSyntax(reg string) bool {
-	return strings.Contains(reg, `(?<`) || strings.Contains(reg, `\1`) ||
+	return strings.Contains(reg, `(?<`) || strings.Contains(reg, `(?=`) || strings.Contains(reg, `\1`) ||
```
bzz: As not everyone may be familiar with intricacies of the regex syntax, would you mind to also updating the GoDoc comment by mentioning the meaning of this new case e.g as "the before text"/lookahead matching?

lafriks: done

PR #237, opened 2019-08-05
Improve shebang detection for osascript when language is specified with -l argument

lafriks:
Part 2 of #234

### DIFF
--- a/common.go
+++ b/common.go
@@ -319,6 +319,12 @@ func getInterpreter(data []byte) (interpreter string) {
 		interpreter = interpreter[:strings.Index(interpreter, `.`)]
 	}
 
+	// If osascript is called with argument -l it could be different language so do not relay on it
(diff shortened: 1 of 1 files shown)


lafriks:
Decrease in coverage is because for this tests are in latest lingust version

common.go
```diff
@@ -319,6 +319,11 @@ func getInterpreter(data []byte) (interpreter string) {
 		interpreter = interpreter[:strings.Index(interpreter, `.`)]
 	}
 
+	// If osascript is called with argument -l it could be different language so do not relay on it
+	if interpreter == "osascript" && bytes.Contains(line, []byte("-l")) {
```
bzz: I think that mentioning that linguist does the same e.g by adding a link to the comment https://github.com/github/linguist/blob/d95bae794576ab0ef2fcb41a39eb61ea5302c5b5/lib/linguist/shebang.rb#L62 will also save some time for the future maintainers.

lafriks: done

Issue #239, opened 2019-08-07
import is a program, not an importable package


4ndre4s:
This is my code: 
```
package main

import (
	"github.com/src-d/enry/cmd/enry"
)

func main() {
	lang, safe := enry.GetLanguageByExtension("foo.go")
}
```
When I run `go get`, I get the following error:
`package github.com/src-d/enry/cmd/enry: import "github.com/src-d/enry/cmd/enry" is a program, not an importable package`

creachadair:
The trouble is that you're importing the command-line tool, rather than the library. Please try using the import path `"github.com/src-d/enry/v2"` instead, e.g., https://play.golang.org/p/qs07S_cvkbT

4ndre4s:
I thought so too, but in the documentary the installation was directly before the use in go, so it seemed plausible to me. Thanks, I'll check tomorrow.

creachadair:
It sounds like we could improve the documentation examples a bit, thanks for pointing that out.

creachadair:
I've updated the README to add some more detail, so I'll close this—but if you still have issues please feel free to re-open it.

PR #240, opened 2019-08-07
Improve download and usage examples in README.md.

creachadair:
Addresses #239. The `go get` command fetches the command-line tool, and does not match the import path for the library. To make things more clear:

1. Mention explicitly that `go get` fetches the CLI. Also, to avoid potential
   issues with pre-modules Go versions, do the fetch in /tmp.

2. Include an import path explicitly in the source examples.

3. Mention explicitly how to import enry into a modules build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,11 @@ File programming language detector and toolbox to ignore binary or vendored file
 Installation
 ------------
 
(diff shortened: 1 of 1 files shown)


creachadair:
Ping @bzz. Not urgent, but it should be a simple review. 😄

PR #244, opened 2019-10-07
docs: Update CLI installation instructions.

### LINKED ISSUE #243: Clarify installation instruction
~~~
(cd /tmp; go get github.com/src-d/enry/v2/cmd/enry)
~~~

https://github.com/src-d/enry#installation

This is misleading, as it assumes the user has already done

~~~
go mod init hello-world
~~~

or similar. If not, it will fail with:

~~~
package github.com/src-d/enry/v2/cmd/enry: cannot find package
"github.com/src-d/enry/v2/cmd/enry"
~~~

creachadair:
Fixes #243. The default behaviour for `go get` has changed slightly and we now need to either provide a module context or disable modules for installation to work correctly. Also remove a now-obsolete reference to the source{d} engine CLI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,13 +24,8 @@ The recommended way to install the `enry` command-line tool is to either
 [download a release](https://github.com/src-d/enry/releases) or run:
 
 ```
-(cd /tmp; go get github.com/src-d/enry/v2/cmd/enry)
+(cd "$(mktemp -d)"; go mod init enry; go get github.com/src-d/enry/v2/cmd/enry)
 ```
-
-This project is now part of [source{d} Engine](https://sourced.tech/engine),
-which provides the simplest way to get started with a single command.
-Visit [sourced.tech/engine](https://sourced.tech/engine) for more information.
-
 
 Examples
 --------
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ The recommended way to install the `enry` command-line tool is to either
 [download a release](https://github.com/src-d/enry/releases) or run:
 
 ```
-(cd "$(mktemp -d)"; go mod init enry; go get github.com/src-d/enry/v2/cmd/enry)
+(cd "$(mktemp -d)" && go mod init enry && go get github.com/src-d/enry/v2/cmd/enry)
 ```
 
 Examples


README.md
```diff
@@ -24,14 +24,9 @@ The recommended way to install the `enry` command-line tool is to either
 [download a release](https://github.com/src-d/enry/releases) or run:
 
 ```
-(cd /tmp; go get github.com/src-d/enry/v2/cmd/enry)
+(cd "$(mktemp -d)"; go mod init enry; go get github.com/src-d/enry/v2/cmd/enry)
```
ncordon: Why not concatenate the commands with `&&` instead of `;`?

creachadair: No good reason, just because it was that way before.

creachadair:
This is a low-risk documentation-only change, so I'm going to merge this once CI passes as @bzz is OOO this week.

cup:
thanks - I would only suggest that

~~~
go mod init enry
~~~

be changed to

~~~
go mod init hello-world
~~~

or similar - as in most cases the user wont be building enry itself, but their own package - and their packages might pull in other packages not in the dependency tree of Enry

creachadair:
> or similar - as in most cases the user wont be building enry itself, but their own package - and their package might pull in other packages not in the dependency tree of Enry

The package name used here is just a nonce, and does not matter: The build happens in a clean temporary directory, so the dependencies of `enry` itself will not affect any other module.

cup:
@creachadair yeah I get that - just syntactically it could be confusing as the word in the position is not required to be (and likely shouldnt be) the same as one of the dependencies.

It would be like me building FFmpeg, and naming the build folder x264. Its not going to hurt anything, its just confusing.

cup:
@creachadair yeah I get that - just syntactically it could be confusing as the
word in the position is not required to be (and likely shouldnt be) the same as
one of the dependencies.

It would be like me building FFmpeg, and naming the build folder x264. Its not
going to hurt anything, its just confusing. Since youre already using `mktemp`,
you could just do something like this:

~~~sh
s1=$(mktemp -d)
s2=$(basename "$s1")
cd "$s1"
go mod init "$s2"
go get github.com/src-d/enry/v2/cmd/enry
~~~

creachadair:
Thanks for the suggestion. I'd prefer to keep this snippet that can be straightforwardly copied and pasted without polluting the current shell environment, however, since that seems to be how people predominantly use quick-install instructions.

Ideally it would just be `go get …` (and in fact it originally was), but the behaviour of the `go` tool varies depending on `$PWD` and the setting of `GO111MODULE`, so we needed to account for that.

I don't really see any potential for confusion here—this snippet is advertised as an alternative way to install the command-line tool, not a usage example. I'd expect anyone doing something more complex than that will understand how to use the `go` tool in a more directed way. I'll defer to @bzz as to whether it's worthwhile to do something more complicated, but I don't really see the need.

PR #247, opened 2019-10-29
doc: cleanup and simplify

bzz:
Make it shorter and more structured, update ToC remove ref from links, etc.


bzz:
Going to address https://github.com/src-d/enry/pull/248#discussion_r340162584 feedback in here

Issue #246, opened 2019-10-15
Language detection accuracy measurements


bzz:
Enry right now consist of the sequence matching of strategies that narrow down the possible language options based on different available information:
 - finelame + extension
 - first line of the content
 - regexp heuristics of the raw content
 - naive bayesian classifier of the tokenized content

As a users, as each strategy can be used independently, I would like to know **how accurate will the language detection be** for each of the distinct use cases.

### Use cases
 - all strategies together (default)
 - filename-only language detection
 - content-only language detection

### Evaluation
Right now, the only measure of overall accuracy of language detection process we have is binary (similar to linguist): if the `linguist/examples/` are all classified or not.

This issue is about picking a better way of quantifying the prediction quality for the three use cases above.

### Steps
 - [ ] identifyling a small dataset, to evaluate up on
 - [ ] a notebook with PoC of the evaluation, to pick the best metric (using Python API from #154)
 - [ ] a script that runs entry for each use case on this dataset \w a given metric  (e.g as part of CI)

The focus of this task is not to get best possible evaluation, but rather to quickly kick off the automation of having at least some evaluation, that will be improved in subsequent work.

Issue #154, opened 2018-08-17
Python bindings for enry


bzz:
Same way as we have Java bindings for enry, wrapping a Go library built with  `-buildmode=c-shared` it would be nice to have one for Python using ~[ctypes FFI](https://docs.python.org/3/library/ctypes.html)~ [cffi](https://cffi.readthedocs.io/en/latest/) or something similar.

Particular use case: one want to use https://github.com/bblfsh/sonar-checks/ API that would require knowing a language the file-to-be-checked is written in, to choose the write checks.

## TODOs
 - [x] Initial PoC: exposes 1-2 API (e.g \wo slices) #245
 - [x] Minimal: expose only high-level language detection API #250
         (usable from Jupiter, after a manual build to enable #246)
 - [ ] Installable: expose _all_ existing API, with the documentation and setup script 
        `pip install -e git+https://github.com/src-d/enry.git#egg=python`
 - [ ] Publish: add new release profile to CI for 
      * building `.whl` (linux, macOS)
      * publishing on pypi

vmarkovtsev:
ctypes is an ancient deprecated way, cool kids use [cffi](https://cffi.readthedocs.io/en/latest/).

bzz:
This is just nice-to-have: for now, the use case listed above is addressed in lookout by incorporating language in DataServer API https://github.com/src-d/lookout/pull/131

bzz:
I've started adding it during OSD and will push simple WIP with CPython extension by cffi soon

bzz:
Issue description updated to include the scope.

PR #250, opened 2019-11-18
python: expose highest-level API

bzz:
All strategies `enry.language()` is now available.

This addresses part of the #154 that should be enough to unblock #246


bzz:
Thank you for the prompt review, @kuba-- 🙇

bzz:
this part is ready to be merged but I do not have enough permissions to do that 😢

PR #251, opened 2020-02-15
Added TSX language color based on Typescript one

Aviortheking:
I added TSX language to the language list in colors example at https://git.delta-wings.net/dzeio/markblog

### DIFF
--- a/data/colors.go
+++ b/data/colors.go
@@ -219,6 +219,7 @@ var LanguagesColor = map[string]string{
 	"Terra":                    "#00004c",
 	"Turing":                   "#cf142b",
 	"TypeScript":               "#2b7489",
+        "TSX":                      "#2b7489",
 	"UnrealScript":             "#a54c4d",
 	"VCL":                      "#148AA8",
 	"VHDL":                     "#adb2cb",


kuba--:
@bzz - ping

bzz:
This file, as anything under `data/*`, is automatically generated from upstream [github/linguist/languages.yml#L4990](https://github.com/github/linguist/blob/d95bae794576ab0ef2fcb41a39eb61ea5302c5b5/lib/linguist/languages.yml#L4990) and thus should not be edited manually as it will be overwritten during the next release process. The change should be applied to the `.yml` file above instead.

bzz:
By looking more at the https://github.com/go-gitea/gitea/pull/10287 I realized there is another option that gitea team might want to consider for color customization, if changes to upstream are not desirable. 

I'm not sure what exactly is the use case here, but as soon as it is exposed on the API surface the client could customize the  `data.LanguagesColor` e.g inside the `init()` function adding whatever is needed, before calling `enry`.

Here is an example: 

```go
package main

import (
	"fmt"

	"github.com/src-d/enry/v2"
	"github.com/src-d/enry/v2/data"
)

func main() {
	fmt.Printf("Before, %d: %s\n", len(data.LanguagesColor), enry.GetColor("TSX"))

	data.LanguagesColor["TSX"] = "#2b7489"

	fmt.Printf("After,  %d: %s\n", len(data.LanguagesColor), enry.GetColor("TSX"))
}
```

Hope this helps!

